PR #8, opened 2024-08-09
Auto elevate with `getKeys`

curtisman:
Refactor the elevate script to be reused by `getKeys` to JIT elevate.

### DIFF
--- a/ts/README.md
+++ b/ts/README.md
@@ -78,9 +78,9 @@ Other examples in the [example directory](./examples/) may have additional servi
 
 Read the [Debugging](#debugging) section for additional service keys that can be used for debugging.
 
-**For AISystems Team Devs**:
+**Local Environment**:
 
-AISystems Team Devs has shared instances of these services. To get the required config and key saved to the `.env` file:
+You can use Azure Key Vault to store keys. To get the required config and keys saved to the `.env` file under the `ts` folder:
 
 - Install [Azure CLI](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli)
 - Run `az login` to login using the CLI. Run `az account set --subscription <Subscription Id>` to set the subscription.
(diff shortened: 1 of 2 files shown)


ts/tools/scripts/elevate.js
```diff
@@ -2,201 +2,21 @@
...
 async function elevate() {
+    const { getClient } = await import("./lib/pimClient.mjs");
+
     // get PIM client
     const client = await getClient();
 
     // elevate
-    var v = await client.elevate();
+    var v = await client.elevate({
+        requestType: config.properties.RequestType,
```
robgruen, line 15: I don't think this will ever be anything other than "SelfActivate" so exposing this seems unnecessary but I wouldn't bother changing it.  Just FYI.

curtisman: Yeah... I just turned any properties that was read from a config before into a parameter.....

ts/tools/scripts/getKeys.mjs
```diff
@@ -2,21 +2,56 @@
...
+        console.warn(chalk.yellowBright("Elevating to get secrets..."));
+        const pimClient = await getPIMClient();
+        await pimClient.elevate({
+            requestType: "SelfActivate",
+            roleName: "Key Vault Administrator",
+            expirationType: "AfterDuration",
+            expirationDuration: "PT5M", // activate for 5 minutes
+        });
+        // Wait for the role to be activated
+        console.warn(chalk.yellowBright("Waiting 5 seconds..."));
```
robgruen, line 41: This elevation API has been known to time out and be throttled on occasion due to heavy usage.  Might want to put retry logic in and wait a bit longer as it's perf is variable depending on the time of day (at least recently).

curtisman: I assume that will just error out and people can just rerun the script?

robgruen: yes, totally reasonable

robgruen: Both actually...YMMV.  I can't say for keyvault as I haven't called that after elevating with the API.

PR #15, opened 2024-08-12
TypeScript root README

curtisman:
- Add a clear overview on how to get started.
- Update `getKey` description (and capabilities for change vault name)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Example agents that use the Microsoft Graph or similar external services may sto
 
 ## Limitations
 
-TypeAgent is early stage sample code over TypeChat. TypeAgent is not a framework. All code in this repo is intented for building our own example apps and agents only.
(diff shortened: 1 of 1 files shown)


ts/README.md
pcdeadeasy, line 1: Suggestion: Right now we are two entry point the cli and shell. One with a light weight GUI experience vs console based experience, can we add something to that effect?

pcdeadeasy, line 1: In my case I had to uninstall the older version of the Azure CLI to get az login to work correctly. Might be worth mentioning that.

curtisman: Sound more like an az cli issue. How about I add latest to the previous line (Install Latest Azure CLI)

curtisman: I think that we should focus the shell on first run. and provide the CLI as an alternative. later.  The CLI is mentioned in the "Running" section later.

curtisman: I have expanded the running section a little bit.

PR #68, opened 2024-08-27
Adding ADO build pipeline YAML

pcdeadeasy:


pipelines/azure-build-ts.yml
```diff
@@ -0,0 +1,83 @@
...
+      npm pack
+    displayName: 'Pack agent-sdk Module'
+
+  - task: npmAuthenticate@0
+    inputs:
+      workingFile: .npmrc
+    displayName: 'Authenticate with Azure Artifacts'
+
+  - script: |
+      npm publish --registry=https://pkgs.dev.azure.com/msctoproj/AI_Systems/_packaging/typeagent/npm/registry/
```
curtisman: can the registory be a variable that get passed in?

pipelines/azure-build-ts.yml
```diff
@@ -0,0 +1,83 @@
...
+  displayName: 'Build TypeScript Project'
+  strategy:
+    matrix:
+      nodeVersion: ['18', '20']
+  steps:
+  - script: |
+      git config --global core.autocrlf false
+    displayName: 'Setup Git LF'
+
+  - script: |
```
curtisman: should we just use the `checkout` task?  Then I don't think you will need the "typeagent_repo" variable

pipelines/azure-build-ts.yml
```diff
@@ -0,0 +1,83 @@
...
+    displayName: 'Setup Node.js'
+    inputs:
+      version: $(nodeVersion)
+      checkLatest: true
+
+  - script: |
+      pnpm install --frozen-lockfile --strict-peer-dependencies
+    displayName: 'Install dependencies'
+    workingDirectory: typeagent/ts
+
```
curtisman, line 63: You will need to change the package number on very build.  Otherwise, it will conflict and won't publish.
I success just append the build number as the prerelease version at the end.   (e.g. 0.0.1-[buildnumber])

pipelines/azure-build-ts.yml
```diff
@@ -0,0 +1,83 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+trigger:
+  branches:
+    include:
+      - main
+
+pr:
+  branches:
```
curtisman, line 10: Probably don't need to build in PR.

pipelines/azure-build-ts.yml
```diff
@@ -0,0 +1,83 @@
...
+    workingDirectory: typeagent/ts
+
+  - script: |
+      cd $(Build.SourcesDirectory)/typeagent/ts/packages/agent-sdk/dist
+      npm pack
+    displayName: 'Pack agent-sdk Module'
+
+  - task: npmAuthenticate@0
+    inputs:
+      workingFile: .npmrc
```
curtisman, line 87: I think you need to create this file?

pipelines/azure-build-ts.yml
```diff
@@ -0,0 +1,83 @@
...
+    displayName: 'Authenticate with Azure Artifacts'
+
+  - script: |
+      npm publish --registry=https://pkgs.dev.azure.com/msctoproj/AI_Systems/_packaging/typeagent/npm/registry/
+    displayName: 'Publish agent-sdk Module'
+    condition: succeeded()
```
curtisman, line 95: You need to specify the tarball file create by `npm pack` to be published.

pipelines/azure-build-ts.yml
```diff
@@ -0,0 +1,85 @@
...
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+variables:
+  node_version: '18'
+  pnpm_version: '9.0.0'
+  typeagent_repo: 'https://github.com/microsoft/typeagent.git'
+  workingDirectory: 'typeagent/ts'
+  ado_registry: 'https://pkgs.dev.azure.com/msctoproj/AI_Systems/_packaging/typeagent/npm/registry/'
```
curtisman: What I mean is that this URL doesn't need to appear in the YML file

pipelines/azure-build-ts.yml
pcdeadeasy: I saw some documentation that this might me created when the pipeline runs?

pipelines/azure-build-ts.yml
pcdeadeasy: So before build change the agentSDK's package.json with an updated version like <current verson>.<build_number>?

pipelines/azure-build-ts.yml
pcdeadeasy: @curtisman I have updated, please check if this is ok.

PR #87, opened 2024-09-01
Improve desktop agent usability.

curtisman:
Usability improvement for the desktop agent.
- desktop app agent automatically start the `autoShell` process without a extra step of `pnpm run start` in the desktop directory.
  - Currently, there is no reason that the desktop controller (`autoShell`) needs to be a singleton process.
  - Remove the websocket shim, and just implement the agent to directly manage it's only `autoShell` process.
- Automatically error and disable desktop agent if `autoShell` isn't built. 
- Save the program name index embedding to be reuse later in the typeagent profile storage. (i.e. shared across sessions), instead of recalculating it on every launch

### DIFF
--- a/ts/packages/agents/desktop/README.md
+++ b/ts/packages/agents/desktop/README.md
@@ -9,18 +9,19 @@ This project has two components
    - Launch Visual Studio
    - Open the repo directory [dotnet/autoShell](../../../../dotnet/autoShell/) folder. Build this project.
 
-2. The node process that integrates with TypeAgent. To build this:
-   - Go to the repo directory [ts/packages/agents/desktop](./) folder.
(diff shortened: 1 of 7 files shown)


hillary-mutisya (approved):
This consolidation/simplification makes sense, now that the desktop agent is running in its own process and can directly spawn autoshell. We probably don't need the agent subfolder under src now - I can take care or moving the files and renaming main.ts after your change goes in.

curtisman:
yeah.  I didn't rename `main.ts` so that the PR can diff better.    Thanks for following up afterward.

If you are going to change the `agent` subfolder, make sure you update export entries in the `package.json` so that the dispatcher can find them.

PR #88, opened 2024-09-03
Remove requestId access in agents

curtisman:
- Browser should use their own id to track calls
- Add `performanceMark` to `ActionContext`.


curtisman:
FYI, @hillary-mutisya and @robgruen

PR #114, opened 2024-09-07
Remove impression interpreter

curtisman:
Remove ImpressionInterpreter because it doesn't work across process boundary when agent is running out of proc.
Since it has minimal use right now, just change it to precompute the value that is needed for the dispatcher (instead of delay calling to interpret). We can design something different if there are additional need.


curtisman:
FYI.

pcdeadeasy (commented):
LGTM

PR #117, opened 2024-09-07
Fix streaming

curtisman:
PR #112  introduce a regression for streaming.   That is because the "content" field of the first chunk is "undefined", and so the code end up returning `<multimodel content>`.   

The non-streaming `complete` function has a laxer requirement for the response, which skips any text before the first "{" and after the last "}" thus avoiding the error (probably a hold over when LLM JSON mode doesn't exist).  The incremental JSON is more strict and bailed on trying parse, and broke streaming.

Looking at both OpenAI and Azure API reference, unlike the input, the response messages are not a complete `PromptSection`. The content for the response can only be string currently. 

The change introduce more accurate typing for the response, and revert the logic change from PR#112

### DIFF
--- a/ts/packages/aiclient/src/openai.ts
+++ b/ts/packages/aiclient/src/openai.ts
@@ -402,6 +402,27 @@ export function supportsStreaming(
     return "completeStream" in model;
 }
 
+// NOTE: these are not complete
+type ChatCompletion = {
+    id: string;
+    choices: {
+        message?: {
+            content?: string | null;
+            role: "assistant";
+        };
+    }[];
+};
+
+type ChatCompletionChunk = {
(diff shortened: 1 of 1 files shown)


curtisman:
FYI, @robgruen 

@hillary-mutisya looks like you made a similar [code change] (https://github.com/microsoft/TypeChat/blame/fb4bdeb8eefc5ed33718c5ed83417f38f83f1480/typescript/src/model.ts#L175) to typechat as well.   You might want to review it to see the impact (of erroring out if `message.content` is `undefine` or `null`)

robgruen:
[like]  Robert Gruen reacted to your message:

PR #131, opened 2024-09-11
Shell agent

curtisman:
- allow the host of dispatcher to inject their own `AppAgentProvider`.
- Create a `shell` app agent and move the shell related commands (and the opaque `shell` settings) to the `shell`
- Refactor the enable/disabled state to the `AppAgentManager`.  The session now stores the "overrides" for enable/disable on session start.


curtisman:
@robgruen Just FYI.

PR #135, opened 2024-09-12
Formalize display format and html and streaming delta

curtisman:
Display format:
- for setDisplay, explicit declare the format is HTML to get correct processing.
- Fix the request message to process the content as HTML (since we switched to getting the content using innerHTML to get images). 
  -  Repros includes if the user request has multiple spaces (which would come out as "&nbsp; ", or any formatting in the input box.
  - Use explicitly format so it can avoid processing pure text as html

HTML output:
- Use dompurify to sanitize HTML given by agent to improve security and prevent XSS.

Streaming delta:
- Instead of replacing the full content of the action during streaming, just send the delta and append.  (In preparation to streaming TTS)
  - incremental JSON parser now sends delta along with the content so far.
  - add `appendDisplay` API.
  - Track streaming context so that we don't replace the output during the actual action execution.

Other improvements:
- player always return TurnImpression, include not found errors.

### DIFF
--- a/ts/packages/shell/src/renderer/src/chatInput.ts
+++ b/ts/packages/shell/src/renderer/src/chatInput.ts
@@ -93,8 +93,9 @@ export function questionInput(
         "replacementDiv",
         "replacement-textarea",
         {
-            onSend: (text) => {
-                chatView.answer(questionId, text, id);
+            onSend: (html) => {
+                // REVIEW: text is from innerHTML, is that ok?
+                chatView.answer(questionId, html, id);
             },
         },
     );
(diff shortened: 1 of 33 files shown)


ts/packages/shell/src/renderer/src/chatInput.ts
```diff
@@ -93,8 +93,9 @@ export function questionInput(
         "replacementDiv",
         "replacement-textarea",
         {
-            onSend: (text) => {
-                chatView.answer(questionId, text, id);
+            onSend: (html) => {
+                // REVIEW: text is from innerHTML, is that ok?
```
curtisman, line 97: @robgruen this change fixed a regression from the vision changes (where the user request display start showing HTML stuff as plain text like &nbsp;) because it didn't expect HTML content that doesn't start with "<".   

However, the `answer` will get HTML content instead.  I added a "REVIEW" here.  just FYI.

PR #139, opened 2024-09-12
Updated dompurify to not strip "target" attributes.

robgruen:
DOMPurify was stripping target attributes so any links in the shell were opening within the shell browser window.


robgruen:
@curtisman FYI

PR #140, opened 2024-09-12
Scroll fix

steveluc:


ts/packages/shell/src/renderer/src/chatView.ts
```diff
@@ -718,9 +718,6 @@ export class ChatView {
         this.topDiv.className = "chat-container";
         this.messageDiv = document.createElement("div");
         this.messageDiv.className = "chat scroll_enabled";
-        this.messageDiv.addEventListener("scroll", () => {
-            this.autoScroll = false;
```
curtisman, line 722: Wouldn't that always scroll to the bottom if there are any content updates.
(That would include any dynamic updates) .

The intent of this code to turn off autoscrolling if people scroll away.  (which will reset to auto scroll when a new request started.)

Is this changed because "scrollIntoView" would raise this event as well.  I am wondering whether there is a replacement.....

steveluc: the sequence is that the event fires for the user request being added, that turns off auto scroll, then assistant message is updated but no scroll happens

we need to distinguish between when a person has scrolled and when scrolling occurs due to overflow

one way to do that might be to check whether scrollHeight === scrollTop 
don't know if that's reliable

steveluc: we resolved this by observing that without alerts there is a small window in which assistant updates can come in

PR #142, opened 2024-09-13
Updated @shell commands

robgruen:
Added all menu items to @ shell command
Fixed a zoom in/out keyboard shortcut bug
Removed tabs


shell:
Confirming that I AM the command 😁

PR #143, opened 2024-09-13
Timing improvements

curtisman:
Improvements
- Add "parse/translation" time and add that to the bottom of the user request message.
- Action time are track individually
- Total time for the whole command is added to the last "agent message"
- First token are now relative to the start of the translate (instead of the last translation from the beginning to request"

Infra:
- Build out the profiling facility to mimic browser/Node.js performance API, with measures/mark, but added facility for hierarchical organization.
- Separate logger (which only create timing entries) to the reader (which create the hierarchical organization for easier interpretation.
- Agent-dispatcher provides interpretation of the measures (via `RequestMetrics`, instead of having the client interpreting the profile data entries)

Future:
- `RequestMetricsManager` will expand to provide summary information
- LLM token data can be store with the `translation` end's metadata.


steveluc:
this is great!

PR #149, opened 2024-09-16
Shell dev UI mode

curtisman:
Hide the perf metrics by default and show only on hover.
Add `devUI` shell settings to allow it always show the perf metrics  (any additional dev time UI can be added under that flag in the future).


curtisman:
FYI @robgruen.   Based on @steveluc feedback, I have hidden the perf metrics by default.

PR #161, opened 2024-09-18
Prevent action context from being used after action is done

curtisman:
Also fix an issue where "translator" overrides get overwritten on load when "undefined" initialize option is passed in.


curtisman:
@pcdeadeasy this should fix the translator setting not being preserved problem

PR #162, opened 2024-09-18
Improve Typing

curtisman:
Improve some of the `any` typing.
Add type on the rpc `invoke` and `call` functions, to ensure the RPC calls are using the right param structures.


curtisman:
@pcdeadeasy Just FYI, this would detect the bug you had fix.

PR #207, opened 2024-10-03
Fix EACCES: permission denied, mkdir '/data' in conversation tests

curtisman:
Use `tmpdir` instead of root of the drive.


curtisman:
@umeshma  FYI.

PR #243, opened 2024-10-11
Agents that supply script with their HTML content are now rendered in sandboxed Iframes

robgruen:
FYI @curtisman

### DIFF
--- a/ts/packages/agentSdk/src/display.ts
+++ b/ts/packages/agentSdk/src/display.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-export type DisplayType = "html" | "text";
+export type DisplayType = "html" | "iframe" | "text";
 
 export type DynamicDisplay = {
     content: DisplayContent;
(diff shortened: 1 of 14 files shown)


curtisman:
Great! Thank you for doing.   

One small change to the API: instead of asking for script to be added, we can change the API for iframe/non-iframe HTML, and just allow non-purified HTML into the iframe?

robgruen:
Good idea, I like it!!!  I'll make that change.

robgruen:
> Great! Thank you for doing this!
> 
> One small change to the API: instead of asking for script to be added, we can change the API for iframe/non-iframe HTML, and just allow non-purified HTML into the iframe? Do you see any potential problems with that?

As for potential problems....I'm sure we'll find out at some point but it does seem reasonable thing to allow.  We can put affordances in the settings or something so people can either turn that functionality on or off so if you're in a spot where you have untrusted agents you can just turn that functionality off.  I'll make a work item for that.

robgruen:
> > Great! Thank you for doing this!
> > One small change to the API: instead of asking for script to be added, we can change the API for iframe/non-iframe HTML, and just allow non-purified HTML into the iframe? Do you see any potential problems with that?
> 
> As for potential problems....I'm sure we'll find out at some point but it does seem reasonable thing to allow. We can put affordances in the settings or something so people can either turn that functionality on or off so if you're in a spot where you have untrusted agents you can just turn that functionality off. I'll make a work item for that.

...I've completed this work but have NOT surfaced the setting in the UI.

PR #258, opened 2024-10-16
Add list commands

steveluc:
Add to coda vs code extension a command to list all of the current commands to a file.


steveluc:
this is probably less useful of a starting point than keybindings.json because the default key bindings probably correspond to commonly used commands

steveluc:
using keybindings.json can also give when clauses, which can help the command implementation to check for conditions (like currently in terminal window) to enable some of the commands; currently no way to get command descriptions from the id strings but there is an open issue for this:
https://github.com/microsoft/vscode/issues/213849

PR #260, opened 2024-10-17
Fix for 'Type quit OR exit to exit' which omits the @ prefix

gvanrossum-ms:
At the start of a session, there's a message `Type quit OR exit to exit`.
However, if you type `quit`, it's an error.
The reason is that this is a command prefixed with `@` (though handled specially in the command parser).

My fix changes this to `Type @quit OR @exit to exit`,
taking the configured command prefix and list of stopcommands into account.

While I was here, I also fixed a missing quote in the final welcome message (`'hello`).

### DIFF
--- a/ts/examples/playground/src/main.ts
+++ b/ts/examples/playground/src/main.ts
@@ -332,7 +332,7 @@ async function runPlayground(): Promise<void> {
         io.writer.writeLine("@help for a list of available commands.");
         io.writer.writeLine();
         io.writer.writeLine(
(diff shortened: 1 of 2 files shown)


gvanrossum-ms:
@microsoft-github-policy-service agree company="Microsoft"

gvanrossum-ms:
@microsoft-github-policy-service agree company="Microsoft"

python.org/~guido

umeshma (approved):
Looks good

PR #295, opened 2024-10-25
Fixed links opening inside of electron window.

robgruen:
_blank links open in the default OS browser


robgruen:
FYI - @hillary-mutisya

PR #313, opened 2024-10-30
First release of Spelunker

gvanrossum:
Right now this mostly exists for debugging the loading and querying of Python code.

The src/main.ts program takes 0 or more .py files on the command line and imports them all into the database (/data/spelunker). The Python code is parsed and split into "chunks" (which may be nested) by a Python script src/chunker.py. (Parsing Python is easy using the `ast` stdlib module; the cost we pay is that we have an silent external dependency on `python3`.)

The Python script writes a JSON list of chunks to stdout which is read by src/pythonChunker.ts and reconstituted into a list of `Chunk` objects in the TS program. (The Python and TypeScript versions of this class and the `Blob` class must be kept in sync manually.)

The src/pythonImporter.ts file then takes this list of chunks and writes them to an `ObjectFolder` (one chunk per folder item), as well as storing an embedding for it using a `SemanticCodeIndex`. Since I had lots of issues here, I verbosely log the comments created by the embedding code interspersed with the source code lines. (This currently shows the documenter is way to eager to create a comment for each line; I'd rather have a higher-level summary for each chunk. Working on that next.)

Once all files have been processed, `main()` then sits in a loop asking the user for a line of text representing a query and looking the nearest N neighbors and shows those hits. (N is currently 5; ideally we should be able reduce that to 1 or 2.) The UI is currently crap.


gvanrossum:
PS. I have no idea what happened to pnpm-lock.yaml. It should only receive a new section for `examples/spelunker:`. and some stuff for `readline-sync` Re-running `pnpm i` doesn't seem to make much of a difference. Maybe I have an unusual version of `pnpm` or something else? Or there's a config file in my `.vscode` that's not under version control?

gvanrossum-ms:
Also note that a few commits (and this PR) are owned by my open-source persona,@gvanrossum, instead of my Microsoft persona @gvanrossum-ms. Does anyone care?

gvanrossum-ms:
HELP! How do I make the error complaining about package.json not being sorted? The sort rules are not obvious.

gvanrossum-ms:
@curtisman Didn't mean for you to review the whole thing (though you're certainly welcome to!). I was just wondering if you could help me with my questions about pnpm-lock.yaml and packages.json.

curtisman:
> @curtisman Didn't mean for you to review the whole thing (though you're certainly welcome to!). I was just wondering if you could help me with my questions about pnpm-lock.yaml and packages.json.

I think running `pnpm run lint:fix` at the root of `ts` directory should fix the problem?

gvanrossum-ms:
> I think running `pnpm run lint:fix` at the root of `ts` directory should fix the problem?

Alas, that doesn't change anything. (It's just an alias for prettier:fix which just calls prettier --write.

I finally looked in the .github/workflows/repo-policy-check.yml file and found the solution: it runs
```
          node ts/tools/scripts/repo-policy-check.mjs
```
and that tool also has a `--fix` flag.

PR #321, opened 2024-10-31
New approach to Spelunker embeddings

gvanrossum-ms:
We create a new code reviewer that produces a single summary paragraph per chunk. We save these summaries in a separate object folder. This is faster and produces more useful embeddings.

Also two speedups:
- Chunkify all files together in a single chunker.py invocation
- Parallelize processing of all chunks for a single file

Finally, some UI improvements: the summaries are not printed during import, but they are printed together on retrieval. This makes importing much quieter now that we've debugged it more.

This required a small interface change to codeIndex.ts: instead of taking a CodeReviewer, createSemanticCodeIndex now takes a narrower interface that only has a document() method.


ts/pnpm-lock.yaml
```diff
@@ -282,7 +285,7 @@ importers:
         version: 18.18.7
       copy-webpack-plugin:
         specifier: ^12.0.1
-        version: 12.0.2(webpack@5.94.0(webpack-cli@5.1.4))
+        version: 12.0.2(webpack@5.94.0)
```
gvanrossum-ms, line 288: @curtisman, @umeshma: Do you understand what these changes are about? I have no idea what pnpm-lock.yaml does. This type of changes is all over the file.

gvanrossum-ms: But I will assume it is somehow correct and I'll still merge this if all tests pass.

gvanrossum-ms: They passed. Merging.

PR #329, opened 2024-11-01
Streaming token count

curtisman:
- Implement streaming token count.
- Track count and max for token count for tags.
- Add type name to the tags for translators.
- Fix counting for tags (it was always counting twice the latest value)
- debug trace now include the translator type name.


curtisman:
@robgruen FYI.

PR #334, opened 2024-11-04
Changed embedding tactics for chunks (again!)

gvanrossum-ms:
New indexing approach:

Instead of asking a documenter to produce a comment for each chunk separately (and without other chunks as context),
ask a different documenter to produce comments for the whole file, but only at points where a chunk starts.
Then extract those comments and add them to the chunk to which they apply.
Use some hackery to pass the relevant comments (with correct line numbers, even!) to the code index
(so it no longer needs an LLM roundtrip to summarize the chunk).

I struggled a bit to get the chunk comments brief, but not too brief --
if you ask for "a paragraph" it explains the `@dataclass` decorator each time it is used. :-)

Also improve the search UX and logging a bit (maybe logging is a bit too verbose, we'll take it down later).

I believe that this improves the quality of the indexing.
However, I don't think it has made much of a difference for the *speed*.
It takes about 13 seconds to document the chunker.py program (nearly 300 lines, 17 chunks).
The embeddings are then super-fast, between 80 and 250 msec per chunk
 I guess this is a *bit* faster -- IIRC the previous approach took about a second per chunk,
but about 4 seconds for the first chunk (for account setup etc.).

### DIFF
--- a/ts/examples/spelunker/src/codeDocSchema.ts
+++ b/ts/examples/spelunker/src/codeDocSchema.ts
@@ -1,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Line of code for which this is a comment
+export type LineDoc = {
+    lineNumber: number;
+    comment: string; // Can be multiline
+};
+
+export type CodeDocumentation = {
+    comments?: LineDoc[];
+};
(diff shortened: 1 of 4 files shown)


gvanrossum-ms:
> I guess this is a _bit_ faster -- IIRC the previous approach took about a second per chunk,
> but about 4 seconds for the first chunk (for account setup etc.).

Hm, no, actually the previous version took about 12 seconds for the same file. (It also produced worse comments, but that doesn't seem to affect the speed.)

PR #355, opened 2024-11-07
Added temporary workaround for streaming token endpoint issue

robgruen:
Turn off streaming token usage if the history contains image_url content.

FYI - @curtisman , @CarlosP-MS


ts/packages/aiclient/src/openai.ts
```diff
@@ -646,11 +648,28 @@ export function createChatModel(
         if (completionSettings) {
             completionParams = { ...completionSettings };
         }
+
+        // BUGBUG - https://learn.microsoft.com/en-us/answers/questions/1805363/azure-openai-streaming-token-usage
+        // image_url content with streaming token usage reporting is currently broken
+        // TODO: remove after API endpoint correctly handles this case
+        let historyIncludesImages: boolean = false;
+        messages.map((ps) => {
+            if (ps.content instanceof Array) {
```
curtisman: Can you use `Array.isArray` instead.   `instanceof` is not always correct.

ts/packages/aiclient/src/openai.ts
```diff
@@ -646,11 +648,28 @@ export function createChatModel(
         if (completionSettings) {
             completionParams = { ...completionSettings };
         }
+
+        // BUGBUG - https://learn.microsoft.com/en-us/answers/questions/1805363/azure-openai-streaming-token-usage
+        // image_url content with streaming token usage reporting is currently broken
+        // TODO: remove after API endpoint correctly handles this case
+        let historyIncludesImages: boolean = false;
+        messages.map((ps) => {
```
curtisman, line 658: You can probably use `message.some` to short circuit in some case?

robgruen: Ah, I didn't know there was such a thing!  I'll take a look.

PR #380, opened 2024-11-13
[spelunker] Add nearestNeighborsPairs to TextTable and createTextIndex

gvanrossum-ms:
This extends the TextTable interface and its two known implementations to support nearestNeighborsPairs, which returns scored pairs of (entry text, array of source ids) for the nearest matches.

I needed this for detailed logging what was going in. (I found that "ast" is somehow a universal attractor in embedding space!)

Also add some clarifying (for me!) comments to createTextIndex (in knowledgeIndex.js) and added some extra (redundant) types that made the meaning of various variables clearer to me (I was mightily confused by the two uses of strings, for entry values and for internal Ids created by the semantic index's put() method).

And some cleanup in spelunker (notably removing a special case to avoid double-scoring exact matches that was based on a misunderstanding) and of course use the new nearestNeighborsPairs for better logging of the index matching mechanism.

I refrained from more serious refactoring.

### DIFF
--- a/ts/packages/knowledgeProcessor/src/knowledgeIndex.ts
+++ b/ts/packages/knowledgeProcessor/src/knowledgeIndex.ts
@@ -84,6 +84,8 @@ export async function createIndexFolder<TValueId>(
     }
 }
 
+export type KVPair<TKey, TValue> = { key: TKey; value: TValue };
+
 export interface TextIndex<TTextId = any, TSourceId = any> {
     text(): IterableIterator<string>;
     ids(): AsyncIterableIterator<TTextId>;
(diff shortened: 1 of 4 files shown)


ts/packages/knowledgeProcessor/src/knowledgeIndex.ts
```diff
@@ -147,6 +154,10 @@ export type TextIndexSettings = {
     embeddingModel?: TextEmbeddingModel | undefined;
 };
 
+// There are *three* important types here:
+// - entries are always strings; typically words or sentences
+// - postings are arrays of TSourceIds; typically unique IDs for other objects
+// - TextId is a string that uniquely identifies an (entry, postings) pair internally
```
gvanrossum-ms, line 158: Is this even right? I am still utterly confused about the meaning of TTextId and TSourceId in TextIndex. Is TTextId there a string that you can feed to an embedding (e.g. a topic or a query) or is it the type of ID generated for this index by the semantic index's put() method? Maybe the naming is just confusing?

PR #447, opened 2024-12-02
Adding support for logging all the LLM exchanges to a MongoDB table. 

CarlosP-MS:
The logged event in the DB contains the prompt text, response text and token usage. It also enables seeing this data as a trace by using a MultiSinkLogger.


ts/packages/aiclient/src/openai.ts
```diff
@@ -518,11 +528,20 @@ function createAzureOpenAIChatModel(
                         if (data.choices && data.choices.length > 0) {
                             const delta = data.choices[0].delta?.content ?? "";
                             if (delta) {
+                                fullResponseText += delta;
                                 yield delta;
                             }
                         }
                         if (data.usage) {
```
robgruen, line 552: If the API doesn't report usage (or usage reporting has been disabled), it's possible that the request won't get logged.  I would move this to the IF above:

                    if (evt.data === "[DONE]") {
                        break;
                    }

CarlosP-MS: Great catch, thank you! I moved the code and confirmed that it works, with and without token usage enabled.

ts/packages/aiclient/src/openai.ts
```diff
@@ -506,23 +514,37 @@ function createAzureOpenAIChatModel(
...
         }
+
+        let fullResponseText = "";
+        let tokenUsage;
         return {
             success: true,
             data: (async function* () {
                 for await (const evt of readServerEventStream(result.data)) {
                     if (evt.data === "[DONE]") {
+                        try {
```
curtisman, line 530: Do you want this to be always logged?  This can be very noisy especially when running tests?  In general, telemetry should be more targeted, no?

robgruen: We could put this behind a setting/telemetry logging level and add an @ command so it can be configured per user?

CarlosP-MS: In the case that triggered the addition of the feature we would like it always logged, but I totally agree that having a setting controlling this is a good idea. I would even make it opt-in once we add it. Can you point me to an existing example for adding a setting and associated command? I can then make it so we can enable/disable at the app level, in addition to letting the user override it.

robgruen: Take a look at **@ config dev on/off** implemented in: https://github.com/microsoft/TypeAgent/blob/main/ts/packages/dispatcher/src/handlers/configCommandHandlers.ts

robgruen: And don't get distracted by the config command handler since that does a lot of stuff.  For a simple command, check out this guy: https://github.com/microsoft/TypeAgent/blob/main/ts/packages/dispatcher/src/handlers/tokenCommandHandler.ts

curtisman: Also note that this library `aiclient` isn't only used in the `shell` or `cli interactive`, it is used in many other cases including examples, test generation, knowledge processing, email processing, calendar indexing, code analysis, etc.   I assume that you want only log a certain use case (or as least able to distinguish the use case so you can able to analyze the data), which mean probably a parameter to pass in a logger (to log to a different table, or add properties to label the user case), might be a better choice here? (as well as the switch to turn it on and off).

robgruen: > Also note that this library `aiclient` isn't only used in the `shell` or `cli interactive`, it is used in many other cases including examples and test generation, etc. I assume that you want only log a certain use case, which mean probably a parameter to pass in a logger might be a better choice here? (as well as the switch to turn it on and off).

Very good point...I do like the "pass in the logger" as a parameter.  If it's not there we don't have to do anything with it... +1

CarlosP-MS: I just pushed a change adding a check for an environment variable that controls whether logging should happen or not, keeping it disabled by default. 

In the end I went with this as a first step because all kinds of requests would be useful for my needs, but if you feel strongly that I should instead condition it only to certain types of model consumption I can move the check for the environment variable to those entry points where the models are created and pass the enable flag as a parameter from there.

While making this change, I also noticed the tags parameter in the CompleteChat requests could be a useful value to add to the events, so I added it. This should help in cases we need to filter out prompt data logging events by area, like it can be done with token counting.

robgruen: This actually meshes well with a change I'm currently working on where I introduce an **@ env** command such that you can see what the current env variables are.

PR #504, opened 2024-12-17
Oracle agent: the "hello world" of agents that can "take over focus"

gvanrossum-ms:
The Oracle agent's is a simple example of an agent that can be the target of the (new, advanced, provisional) `@` command
```
@config request <agent>
```
The requirements for such an agent are:

- Must support commands (not just actions)
- Must have a command named `request` with a single param of type `string` that has `implicitQuotes` set

This particular agent also supports an action, named `queryOracle`, with a string param `query`. That action has the same effect as the `request` command: Because this is a hello-world app, both just return a random string (in effect, a fortune cookie). The control flow for actions and commands is quite different, as is the way their schemas are described, so this is a useful comparison. The actual random-string-returning part is factored out in the `makeRandomResponse()` function, called from both action and command.


ts/packages/agents/oracle/src/oracleActionHandler.ts
```diff
@@ -0,0 +1,140 @@
...
+export function instantiate(): AppAgent {
+    return {
+        initializeAgentContext: initializeOracleContext,
+        executeAction: executeOracleAction,
+        ...getCommandInterface(handlers),
+    };
+}
+
+type OracleActionContext = {
+    enabled: boolean;
```
gvanrossum-ms: This is a vestige of an earlier design. I would take it out but right now I can't even push to my own fork due to auth issues on my new laptop. :-(

gvanrossum-ms:
```suggestion
```

ts/packages/agents/oracle/src/oracleActionHandler.ts
```diff
@@ -0,0 +1,140 @@
...
+};
+
+async function initializeOracleContext(): Promise<OracleActionContext> {
+    return { enabled: true };
```
gvanrossum-ms:
```suggestion
    return { };
```

PR #505, opened 2024-12-17
MS Graph login flow and fixes

curtisman:
- Calendar and email: default MS Graph login on start up or triggering action to silent mode (i.e. don't prompt user). 
  - If it cannot authenticate, then return error message to use `@calendar login` or `@email login` to explicit trigger the device sign in prompt.
  - With the explicit login commands, display the device login flow prompt on the chat view, and wait for login to complete.
- Add explicit `@calendar logout` and `@email logout` to explicit switch to auth state.  

- Background indexCalendarThread: properly track the setInterval return so that it can actually stop
- Use `proper-lock` library to use the fs to prevent multiple silent flow log in attempt.
- Switch `console.log` to use the `debug` package in calendar/email and graph util
- Clean up `GraphClient` API, remove unused code.   Also no need to track access token expiry.  The azure library takes cares for that.
- Fix spelling of variable `calendatDataIndex` => `calendarDataIndex`

- Removed "fuzzy response" in  `calendar.addEvent`, instead rely on clarify

### DIFF
--- a/ts/packages/agents/agentUtils/graphUtils/src/calendarClient.ts
+++ b/ts/packages/agents/agentUtils/graphUtils/src/calendarClient.ts
@@ -5,6 +5,7 @@ import { GraphClient, DynamicObject, ErrorResponse } from "./graphClient.js";
 import registerDebug from "debug";
 import chalk from "chalk";
 import { createCalendarDataIndex } from "./calendarDataIndex.js";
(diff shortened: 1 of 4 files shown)


curtisman:
@pcdeadeasy FYI.

pcdeadeasy:
Thanks for the fixups @curtisman.

PR #508, opened 2024-12-17
Added action name (when available) to agent message timestamp label 

robgruen:
- fixed randomcommand
- added DEVUI checkbox to the settings dialog

image


ts/packages/dispatcher/src/execute/actionHandlers.ts
```diff
@@ -190,6 +190,9 @@ async function executeAction(
     // Update the last action translator.
     systemContext.lastActionSchemaName = translatorName;
 
+    // Update the last action name.
+    systemContext.lastActionName = action.fullActionName;
```
curtisman, line 194: Might want to set it back to undefined int he finally below after execution?

robgruen:
Yeah, good point...will do.

ts/packages/dispatcher/src/execute/actionHandlers.ts
robgruen: good point, will do.

PR #511, opened 2024-12-19
Disable oracle schema by default

curtisman:
Oracle is in conflict with the chat agent.   Disable the schema by default.

Also rename the setting to `schemaDefaultEnabled` to align terminology


curtisman:
@steveluc @gvanrossum-ms FYI.

PR #534, opened 2025-01-08
Remove function to create lock file when tokencache.bin is missing.

pcdeadeasy:
Instead use withLockFile on the directory name.


pcdeadeasy:
@curtisman FYI.

PR #539, opened 2025-01-09
API Storage Provider Support

hlucco:
Abstracted the Azure Blob storage code in `api` into `TypeAgentStorageProvider` to support using s3 or any other storage provider for session storage as well.

Note: I do not have the ability to test the Azure Storage Provider, would be prudent to have someone with azure credentials do a test run with blob storage to ensure it is working as expected.


robgruen:
I can confirm that this works with the azure storage provider, nice work!

PR #542, opened 2025-01-10
Prompt rebalancing

curtisman:
Adjust the comment for `chat.generateResponse` to make the scenario that explain common phrase the user type more stable.
Add prompt instruction to avoid clarifying unresolve resolution if there is a possible match (which also currently stop it from activate if there are multiple ambiguity. Will continue to rebalance over time).


steveluc:
💯

PR #565, opened 2025-01-17
Implement RPC for ClientIO interface

curtisman:
Switch electron to use it.
Also switch the web api to use the Dispatcher and ClientIO RPC, which means most of the placeholder are gone.


ts/packages/api/src/webSocketServer.ts
robgruen, line 1: Much simpler...I like it!

PR #582, opened 2025-01-22
Make spelunker's dependency better-sqlite3 work in both electron and node context.

curtisman:
Make a copy of the node version and explicitly pass the path to `new Database` when running in node context.


curtisman:
@gvanrossum-ms can you test your spelunker scenario for in both `shell` and `cli`

PR #585, opened 2025-01-22
Remove tsc compiler reference same source twice

curtisman:
testData.ts don't need to be rebuild in the test directory in memory provider and knowledge processor.
It interferes with incremental building.


curtisman:
@umeshma FYI

PR #596, opened 2025-01-25
Shell init order

curtisman:
Make sure the DOM is ready, where both side of the RPC is setup, before using the dispatcher.

This fix the issue where the greeting sometime is too early, calling the chatView before it is initialized


curtisman:
@robgruen FYI.

robgruen:
> @robgruen FYI.

I fixed this earlier today in my test branch!  Good find!

PR #598, opened 2025-01-26
Fix vscode agent hang

curtisman:
Fix regression from #583 

The `data.id` is actually a number in this case (which is just round trip from the agent), and a plain "truthy" check would reject call id of 0, causing the remote call to not be resolved.

Do an explicit `undefined` check instead.


curtisman:
@hillary-mutisya FYI.

PR #602, opened 2025-01-27
Fix graph client login regression

pcdeadeasy:
- Change version of `@azure/identity` to version 4.2.1 instead of 4.5.0.
- The issue with version 4.5.0 is documented in AzureAD [issues](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/7303)
- Without this fix the token is not being read from the cache and we need to login every single time event. (Even when the token is valid and does not need to be refreshed.)


pcdeadeasy:
fyi: @robgruen and @curtisman

curtisman (approved):

PR #607, opened 2025-01-28
Bug fix when handling promises.

pcdeadeasy:
- Was using Promise.race which causes non-deterministic behavior. The code path uses separate async function to handle actions, updated to Promise.all.
- Fixed code handle actions objects where parameter property is undefined.


pcdeadeasy:
FYI @curtisman

PR #637, opened 2025-01-30
Switch to use `WebContentsView` from deprecated `BrowserView`

curtisman:
Also fix the app path in the playwright tests for windows to have `file://` prefix.


curtisman:
@hillary-mutisya FYI.

hillary-mutisya:
looks great!

PR #781, opened 2025-03-04
Added setDisplayInfo() to RPC so that agent command handlers can call it

robgruen:
CommandHandlers can now call setDisplayInfo() to report action-data for debugging purposes.

image

### DIFF
--- a/ts/packages/agents/spelunker/evals/eval-1/source/dispatcher/src/context/commandHandlerContext.ts
+++ b/ts/packages/agents/spelunker/evals/eval-1/source/dispatcher/src/context/commandHandlerContext.ts
@@ -10,7 +10,7 @@ import {
     MultiSinkLogger,
     createDebugLoggerSink,
     createMongoDBLoggerSink,
-    Profiler
+    Profiler,
 } from "telemetry";
 import { AgentCache } from "agent-cache";
 import { randomUUID } from "crypto";
--- a/ts/packages/agents/spelunker/evals/eval-1/source/dispatcher/src/context/commandHandlerContext.ts
+++ b/ts/packages/agents/spelunker/evals/eval-1/source/dispatcher/src/context/commandHandlerContext.ts
@@ -10,6 +10,7 @@ import {
     MultiSinkLogger,
     createDebugLoggerSink,
     createMongoDBLoggerSink,
+    Profiler
 } from "telemetry";
 import { AgentCache } from "agent-cache";
 import { randomUUID } from "crypto";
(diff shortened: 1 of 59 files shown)


robgruen:
@curtisman, could you please take a look at this and make sure that I've put the requestId where it should go based on your design pattern. Let me know if you want me to refactor.

ts/packages/agents/spelunker/evals/eval-1/source/dispatcher/src/context/commandHandlerContext.ts
```diff
@@ -24,16 +25,15 @@ import { TypeAgentTranslator } from "../translation/agentTranslators.js";
 import { ActionConfigProvider } from "../translation/actionConfigProvider.js";
 import { getCacheFactory } from "../utils/cacheFactory.js";
 import { createServiceHost } from "./system/handlers/serviceHost/serviceHostCommandHandler.js";
-import { ClientIO, RequestId, nullClientIO } from "./interactiveIO.js";
+import { ClientIO, nullClientIO } from "./interactiveIO.js";
```
curtisman, line 28: @robgruen I doubt you want to change this file.  This is a fixed cache of the source for spelunker to consume.
@gvanrossum-ms is a script to check out based on the hash still in the plan so we can remove this from the repo.

gvanrossum-ms: Not exactly (it exceeds my git fu a little bit). How about I change all the file extensions to `.py.txt`? That should prevent mistakes like this, and is easy to compensate for in the spelunker code.

robgruen: If I don't modify it, the project won't build...I could add a duplicate type declaration so that this still builds but that seems wrong.  Thoughts?

robgruen: > Not exactly (it exceeds my git fu a little bit). How about I change all the file extensions to `.py.txt`? That should prevent mistakes like this, and is easy to compensate for in the spelunker code.

works for me...

robgruen: Another option is to exclude this folder from the build.

robgruen: Will rework and revisit shortly.

PR #784, opened 2025-03-04
Added appendDiagnosticData() to clientIO interface 

robgruen:


robgruen:
FYI @curtisman

ts/packages/shell/src/renderer/src/main.ts
```diff
@@ -80,6 +80,9 @@ function addEvents(
         appendDisplay: (message, mode) => {
             chatView.addAgentMessage(message, { appendMode: mode });
         },
+        appendDiagnosticData: (requestId, data) => {
+            chatView.setActionData(requestId, data);
```
curtisman, line 84: Might want to leave a TODO comment about doing append later?

curtisman: Although you could have implemented append easily by storing that separately in an array that always display after the "action"

robgruen:
I did but it's in messagecontainer.ts.  I'll put one here as well!

PR #923, opened 2025-04-10
Refactor Client API and others

curtisman:
- Instead of register handler for each main->renderer call individually, set it up with one "register call" in the reload script.
- Refer to settings in the shellWindow instead of getting it thru ShellSettings.getinstance() when possible.
- Fix greeting agent that if the call the Azure CLI failed to get the current user name, it should still resolve the promise.
- Change the chat history from pushing from main to render, to instead have the render call the main process to get the history.
- The full electronAPI is no longer exposed in the chat view.


curtisman:
@robgruen FYI on the greeting agent fix.

PR #1006, opened 2025-05-01
Fix test instability

curtisman:
Get occasional failure in `sqliteCommon.ts` createDatabase function in test complaining unable to create database because the path doesn't exist.  That's because in `ensureDir`, it calls the promise version of `mkdir` without await.  So it doesn't guarantee that the directory is actually created before it returns.   Switch to the `sync` version


curtisman:
@umeshma, I assume that is the intent?

PR #1140, opened 2025-05-17
Fix inline browser extension message round trip for WebAgent 

curtisman:
- Fix two issues with the message with the result from the dispatcher to the browser doesn't get thru because
  - The data sometimes are blobs.
  - The Electron IPC channel name is `"received-from-browser-ipc` instead of webView
- Add turtle graphics as one of the sites for shell open action.
- Fix spelling.


curtisman:
@hillary-mutisya can you please make sure these are correct.  If not, I can back them out.

PR #1157, opened 2025-05-20
Azure ARM template

curtisman:
- Make the azure ARM template work with just the command `az deployment sub create --template-file template.json --location eastus` after logged in to AzCLI and chosen the default subscription.
  - Remove non-essential services (mongodb and storage accounts for the docker service, will add those separately later if needed)
  - Reduce capacity to reasonable level for trying things out to fit the default quota.
  - Change it to deploy to a subscription and automatically create everything, including RBAC role assignments, and keyvault value.
 - TypeAgent will assume authentication using "identity" when key is missing.

Remaining work not in this PR:
- Need to switch our search API/implementation because Bing search API is going away.
- The key vault is using the same name per region, so will conflict if there are multiple people using the default value.   Consider skipping the key vault and just generating the .env file, or generate a unique name and provide it to the user.

### DIFF
--- a/azure/parameters.json
+++ b/azure/parameters.json
@@ -1,21 +1,45 @@
 {
-    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
-    "contentVersion": "1.0.0.0",
-    "parameters": {
-        "accounts_agentmaps_name": {
-            "value": null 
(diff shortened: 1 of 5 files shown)


curtisman:
@robgruen do you mind reviewing this and testing it?

PR #1216, opened 2025-06-13
Improve unknown page open experience.

curtisman:
- Add comment in the action schema to push the model to not translate the URL based on trained data, which could be outdated, and allow the agent to resolve the URL based on current grounded information instead  (also added test)
- Test if the input for site is already an URL and directly return it.
- Restore the try/catch around the local port resolution so that it doesn't throw if the input is not an agent name
- Add status message when resolving entities (which can take a while for web pages)


robgruen:
Thanks for adding some tests, you beat me to it!

PR #1254, opened 2025-06-24
Added urlResolver and urlValidate commands to schema studio

robgruen:
These two commands ensure that URLs can and are resolve properly in support of the OpenWebPage action of the browser agent.

Added azure-ai-foundry package that wraps up all calls to azure agent/project SDKs.


ts/package.json
```diff
@@ -51,7 +51,7 @@
     "prettier": "^3.5.3",
     "shx": "^0.4.0"
   },
-  "packageManager": "pnpm@10.12.1+sha512.f0dda8580f0ee9481c5c79a1d927b9164f2c478e90992ad268bbb2465a736984391d6333d2c327913578b2804af33474ca554ba29c04a8b13060a717675ae3ac",
+  "packageManager": "pnpm@10.12.2",
```
curtisman: Can you use `corepack up` to update this so that we have the hash?

PR #1278, opened 2025-06-27
Update version for vite and mermaid

curtisman:


curtisman:
@hillary-mutisya  FYI

PR #1431, opened 2025-08-07
Added reentrancy gates when websocket is created/dropped 

robgruen:
If the browser extension is not installed there's a potential timing issue that causes the machine to go into a create/close WS loop
- now there's a singular Promise while waiting for port creation
- dropped connections are now throttled


robgruen:
@hillary-mutisya , would you mind giving a once over please just to make sure.  I checked and both the external and internal browser commands appear to work for me.

Issue #1466, opened 2025-08-19
cache works in browser activity


steveluc:
Cache continues to work even in a focused activity.

PR #1473, opened 2025-08-20
Allow cache behavior to be specified during activity.

curtisman:
For now, add capability to enable caching during activity.
- By default, use of cache is disabled during activity.   
- Add property under `cachedActivites` in the agent manifest to enable cache.   
  - Activity name is used as the property name.
  - If the value is true then cache is enable (and shared with non-activity cache).
  - If the value is `separate` then it will have a separate namespace in the cache. for the activity.
- Enable matching using activity app agent's cache (separate or shared based on config).
  - PENDING: Second chance translation (with non-activity app agent) doesn't use cache yet.
- Completion will resolve completion for both activity (based on config) and non-activity cache.  (And will disable completion for the activity agent if config say to not cache)

### DIFF
--- a/ts/packages/agentSdk/src/agentInterface.ts
+++ b/ts/packages/agentSdk/src/agentInterface.ts
@@ -18,9 +18,10 @@ export type IndexingServiceConfig = {
 
 export type IndexingServicesManifest = Record<string, IndexingServiceConfig>;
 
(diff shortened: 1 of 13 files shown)


curtisman:
Part 1 for #1466

PR #1478, opened 2025-08-20
Return undefined instead of invalid action names for partial matches 

### LINKED ISSUE #1474: No completion if the cache has action name that are partially matched
Repro:
- Populate the cache with `open bbc webpage`, where the explainer will indentify "open" and "webpage" as subphrase that translate to "browser.open".
- Then completion will fail with `open` because it couldn't transform the action name with only the first part of the subphrase.

lmcrean:
Partially fixes #1474 - Prevents exception when typing partial cached commands by properly handling undefined action names.

### Problem
Cache contains "open bbc webpage" → "browser.openWebPage". Typing just "open" throws exception because:
1. `constructionValue.ts` returns `{ fullActionName: "unknown.unknown" }` 
2. `fromJsonAction` tries to parse "unknown.unknown" and fails

### Solution
Two coordinated changes to handle partial matches gracefully:

**1. constructionValue.ts**: Return `undefined` instead of "unknown.unknown" for partial matches
```diff
- return { fullActionName: "unknown.unknown" };
+ return { fullActionName: undefined }; // Return undefined for partial matches
```

**2. requestAction.ts**: Handle undefined `fullActionName` in `fromJsonAction`
```typescript
const { schemaName, actionName } =
    actionJSON.fullActionName !== undefined
        ? parseFullActionNameParts(actionJSON.fullActionName)
        : { schemaName: undefined as any, actionName: undefined as any };
```

### Testing
1. Execute: `open bbc webpage` (populates cache)  
2. Type: `open` (partial match triggers completion)
3. **Before**: Throws exception attempting to parse "unknown.unknown"
4. **After**: No exception - returns undefined gracefully, preventing crash


curtisman (changes requested):
PR implement completion behavior we are looking for.   Then completion should include a list of websites instead of the merge of two separate subphrase in the construction.

ts/packages/cache/src/constructions/constructionValue.ts
curtisman, line 1: Code around line 187 `return { fullActionName: "unknown.unknown" };` needs to be changed as well.

ts/packages/cache/src/explanation/requestAction.ts
```diff
@@ -194,6 +194,25 @@ function executableActionsToString(actions: ExecutableAction[]): string {
 }
 
 function fromJsonAction(actionJSON: JSONAction) {
+    // Handle undefined fullActionName for partial matches
```
curtisman: Note to myself: I don't like how this validate the typing (casting `undefined` as `strings`) that might expose problems down the line.  But fixing the typing will probably require more extensive change then warranted for this fix.  I will create a separate issue to track.

ts/packages/cache/src/explanation/requestAction.ts
```diff
@@ -194,6 +194,25 @@ function executableActionsToString(actions: ExecutableAction[]): string {
...
+        const executableAction: ExecutableAction = {
+            action,
+        };
+        if (actionJSON.resultEntityId !== undefined) {
+            executableAction.resultEntityId = actionJSON.resultEntityId;
+        }
+        return executableAction;
+    }
+    
     const { schemaName, actionName } = parseFullActionNameParts(
```
curtisman: Instead of duplicating the code from `createExecutableAction` above, we can just cast the value for schemaName and actionName, i.e.
```typescript
    const { schemaName, actionName } =
        actionJSON.fullActionName !== undefined
            ? parseFullActionNameParts(actionJSON.fullActionName)
            : { schemaName: undefined as any, actionName: undefined as any };
```

ts/packages/cache/src/explanation/requestAction.ts
lmcrean: -- Noted

ts/packages/cache/src/explanation/requestAction.ts
lmcrean: thanks -- perhaps a seperate ticket for this could be created if hasn't already (note to self).

lmcrean:
@curtisman thank you for your technical notes, I sense a partial fix is optimal at this point and have narrowed the scope

ts/packages/cache/src/explanation/requestAction.ts
curtisman: The change to handle undefined "fullActionName" will still be needed?  Otherwise, it would throw in `parseFullActionNameParts`, no?

lmcrean: yes you're right, I have updated

curtisman:
@lmcrean The PR is failing lint check currently.  You can run `pnpm run lint:fix` to auto fix them.

PR #1484, opened 2025-08-20
Refactor confirm and logging code from `matchRequest` and `translateRequest`

curtisman:


curtisman:
Prepare for part 2 for #1466

Issue #1483, opened 2025-08-20
Shell: Browser action hang when no inline browser open.


curtisman:
e.g. without inline browser open, and request "scroll up browser" result in the following error.

Console error print out.
```
Failed to send RPC message to browser: Error: No active browser tab available
    at getActiveBrowserWebContents (file:///home/curtism/src/TypeAgent/ts/packages/shell/out/main/index.js:2407:13)
    at file:///home/curtism/src/TypeAgent/ts/packages/shell/out/main/index.js:2413:27
    at Object.send (file:///home/curtism/src/TypeAgent/ts/packages/agentRpc/dist/common.js:31:13)
    at out (file:///home/curtism/src/TypeAgent/ts/packages/agentRpc/dist/rpc.js:12:17)
    at file:///home/curtism/src/TypeAgent/ts/packages/agentRpc/dist/rpc.js:94:17
    at new Promise (<anonymous>)
    at Object.invoke (file:///home/curtism/src/TypeAgent/ts/packages/agentRpc/dist/rpc.js:93:20)
    at Object.scrollUp (file:///home/curtism/src/TypeAgent/ts/packages/agents/browser/dist/common/contentScriptRpc/client.mjs:7:48)
    at Object.scrollUp (file:///home/curtism/src/TypeAgent/ts/packages/shell/out/main/index.js:2464:35)
    at Object.executeBrowserAction [as executeAction] (file:///home/curtism/src/TypeAgent/ts/packages/agents/browser/dist/agent/actionHandler.mjs:692:60)
``

curtisman:
@hillary-mutisya FYI.

hillary-mutisya:
the error message seems correct (the system correctly detected that there is not active browser tab). This should show a message to the user and not cause an action hang.

PR #1491, opened 2025-08-20
Fix: Shell: Browser action hang when no inline browser open

hillary-mutisya:
- FIx https://github.com/microsoft/TypeAgent/issues/1483
     - Propagate error messages from inlinebrowsercontrol and show the message to the user.
- Fix codescan warning https://github.com/microsoft/TypeAgent/security/code-scanning/181


hillary-mutisya:
@curtisman  FYI

PR #1523, opened 2025-08-29
Improve unknown action comment

curtisman:
Improve the pass rate of the gate.


ts/packages/dispatcher/src/context/dispatcher/schema/dispatcherActionSchema.ts
```diff
@@ -3,7 +3,7 @@
 
 export type DispatcherActions = UnknownAction;
 
-// user request isn't any of the action available and no schema group that may have potential actions to look up from.
+// Use UnknownAction when the action available in the schema is relevant to the user request
```
robgruen: Should this read:

// Use UnknownAction when the action available in the schema is **not** relevant to the user request.

Or did I read it wrong?

curtisman: Yep.  That was wrong.

PR #1568, opened 2025-09-11
[python/knowpro] Omnibus PR

gvanrossum-ms:
- Make load_dotenv() also look in ./.env, ../.env, ../../.env, all the way up till /.env
- Misc minor cleanups
- Remove out-of-spec get_related_terms() methods
- Rename _terms_to_ordinal to _added terms in SqliteRelatedTermsFuzzy
- Disable remove_term() -- it cannot work for now
- Remove lookup fallback function
- Simplify and correct RelatedTermsFuzzy table to just (term, embedding)
-

### DIFF
--- a/python/ta/TODO.md
+++ b/python/ta/TODO.md
@@ -2,6 +2,7 @@
 
 # TODOs for new repo setup
 
+- Merge newer changes from TypeAgent repo
 - Vendor TypeChat (Python version)
 - Update load_dotenv() to look for .env in current directory and going up (*plus* ts/.env)
 


python/ta/test/test_sqlite_indexes.py
```diff
@@ -28,18 +28,10 @@
     SqliteRelatedTermsIndex,
 )
 from typeagent.storage.sqlite.schema import init_db_schema
+from typeagent.storage.sqlite.messageindex import SqliteMessageTextIndex
+from typeagent.knowpro.convsettings import MessageTextIndexSettings
```
robgruen: nit: does it make sense to group with other typeagent.knowpro imports above?  You're the expert, but it seems like a good place to put it.

python/ta/test/test_sqlite_indexes.py
```diff
@@ -491,3 +461,195 @@ async def test_fuzzy_index_first_run_scenario(
...
+
+class TestSqliteIndexesEdgeCases:
+    """Test edge cases and error conditions in SQLite indexes."""
+
+    @pytest.mark.asyncio
+    async def test_term_index_edge_cases(self, sqlite_db: sqlite3.Connection):
+        """Test edge cases in term index."""
+        index = SqliteTermToSemanticRefIndex(sqlite_db)
+
+        # Test with None/empty lookups
```
robgruen, line 482: Do we do any tests anywhere with white space, unicode, or other punctuation type of queries?

python/ta/test/test_sqlite_indexes.py
```diff
@@ -491,3 +461,195 @@ async def test_fuzzy_index_first_run_scenario(
...
+        """Test edge cases in property index."""
+        index = SqlitePropertyIndex(sqlite_db)
+
+        # Test lookup of non-existent property
+        results = await index.lookup_property("nonexistent", "value")
+        assert results is None
+
+        # Test removal operations
+        await index.add_property("test_prop", "test_value", 1)
+        await index.remove_property("test_prop", 1)
```
robgruen, line 515: I would add an assert here to make sure it was added correctly.  I could envision a scenario where the test passes but the functionality is broken because in fact the "add_property" never added it.

robgruen: Same for the add_terms from the previous test.

python/ta/test/test_sqlite_indexes.py
gvanrossum-ms: Yes, will do. I am actually pretty OCD about import order but Claude needs a little more help.

python/ta/test/test_sqlite_indexes.py
gvanrossum-ms: Good idea. I will ask Claude to add some of those, but not tonight.

python/ta/test/test_sqlite_indexes.py
gvanrossum-ms: Fixed x2.

python/ta/typeagent/aitools/utils.py
```diff
@@ -71,14 +71,28 @@ def reindent(text: str) -> str:
 
 def load_dotenv() -> None:
     """Load environment variables from '<repo_root>/ta/.env'."""
-    dn = os.path.dirname
-    repo_root = dn(dn(dn(dn(dn(__file__)))))  # python/ta/typeagent/aitools/utils.py
+    paths = []
+    # Look for <reporoot>/ts/.env first.
+    repo_root = os.popen("git rev-parse --show-toplevel").read().strip()
```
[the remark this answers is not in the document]
gvanrossum-ms: Fixed. I added this to be prepared for when the code is moved to /microsoft/typeagent-py. I have changed it so that if `repo_root` is None it will not attempt to calculate `<repo_root>/ts/.env` nor add it to `paths`. popen() will let stderr from the command go to stderr of the Python process; I'll watch for whether that's useful or not and then decide if it needs to be redone more.

gvanrossum-ms (commented):
Fixed the nits, will add more edge cases later.

PR #1616, opened 2025-09-19
Add button to toggle shell layout

hillary-mutisya:


ts/packages/shell/src/main/shellWindow.ts
```diff
@@ -507,6 +508,7 @@ export class ShellWindow {
 
         const verticalLayout = this.verticalLayout;
         const { width, height } = bounds;
+        const availableHeight = height - TITLE_BAR_HEIGHT; // Account for title bar
```
curtisman, line 511: I am wondering whether this is still necessary after my recent fix  in PR #1614. Where I set the position and content sizes separately instead of the bounds.

PR #1647, opened 2025-09-24
Shell: fix exception on title update after window is closed

curtisman:


curtisman:
@robgruen just noticed that this will conflict with your PR #1644

robgruen:
> @robgruen just noticed that this will conflict with your PR #1644

Mine also has a minor update to show the title but I don't handle the exception.  Want me to incorporate that into mine?

PR #1923, opened 2026-02-15
changes for spotify since feb 2026:

bmerkle:
- limit parameter has changed from 50 to 10 (otherwise we get API error now)
- search for music titles has been improved to contain also special characters.

Note:
- a spotify premium account is necessary to use the spoifty via API
- therefore it is good, that we have localPlayer as alternative to spoify player

### DIFF
--- a/ts/examples/memoryProviders/package.json
+++ b/ts/examples/memoryProviders/package.json
@@ -28,7 +28,7 @@
     "tsc": "tsc -b"
   },
   "dependencies": {
-    "@elastic/elasticsearch": "^8.17.0",
+    "@elastic/elasticsearch": "^8.19.1",
     "aiclient": "workspace:*",
     "better-sqlite3": "12.2.0",
     "debug": "^4.4.0",
--- a/ts/package.json
+++ b/ts/package.json
@@ -54,6 +54,7 @@
     "@fluidframework/build-tools": "^0.57.0",
     "@types/node": "^20.17.28",
     "markdown-link-check": "^3.14.2",
+    "prebuild-install": "^7.1.3",
     "prettier": "^3.5.3",
     "shx": "^0.4.0"
   },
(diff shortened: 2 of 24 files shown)


ts/packages/agents/player/src/endpoints.ts
```diff
@@ -373,6 +373,7 @@ export async function getArtistTopTracks(service: SpotifyService, id: string) {
     return fetchGet<SpotifyApi.ArtistsTopTracksResponse>(
         service,
         `https://api.spotify.com/v1/artists/${encodeURIComponent(id)}/top-tracks`,
+        { market: "US" },
```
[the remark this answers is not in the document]
bmerkle: this is only a default

ts/packages/agents/player/src/client.ts
```diff
@@ -514,7 +514,7 @@ export async function searchTracks(
     const query: SpotifyApi.SearchForItemParameterObject = {
         q: queryString,
         type: "track",
-        limit: 50,
+        limit: 10,
```
robgruen: Let's make this a constant and use it everywhere you changed it (or, make a default settings that has this value and update the current call sites to use and merge with whatever setting needs non-default values.

ts/packages/agents/player/src/endpoints.ts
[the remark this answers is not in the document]
robgruen: I'm kinda with co-pilot on this one.  Do we need this, how does it affect the app results?  If we do add this, we shouls make it modifiable with a command and save it with theb player settings.

ts/packages/agents/player/src/client.ts
bmerkle: fixed

ts/packages/agents/player/src/endpoints.ts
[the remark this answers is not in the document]
bmerkle: removed

ts/packages/agents/player/src/client.ts
robgruen: Can we please reuse these constants in ts/packages/agents/player/src/search.ts as well.

bmerkle: fixed (also in additional places)

PR #1953, opened 2026-02-26
Added OOBE key check on first run

### LINKED ISSUE #1892: Need OOBE experience
- on first run, need to check keys, etc.

avdeshjadon:
added a check on first run to see if the required api keys are set. if they're missing, it shows a dialog with what keys are needed and a button to open the setup docs.

Closes #1892


ts/packages/shell/src/main/keys.ts
```diff
@@ -137,6 +137,43 @@ export async function loadKeysFromEnvFile(envFile: string) {
...
+            "No valid service keys were found. Please configure at least one of:\n\n" +
+            "• Azure OpenAI: AZURE_OPENAI_API_KEY + AZURE_OPENAI_ENDPOINT\n" +
+            "• OpenAI: OPENAI_API_KEY + OPENAI_ENDPOINT\n\n" +
+            "Click 'Open Setup Docs' to learn how to set up your keys.",
+    });
+
+    if (result.response === 0) {
+        shell.openExternal(
+            "https://github.com/microsoft/TypeAgent/blob/main/ts/README.md",
+        );
```
robgruen, line 173: Let's exit the application after the user selects "open external."  When I did the main app continued to execute and throw an error which was a bit unexpected and a lesser user experience.  If the user hits "continue anyway" I think it's ok to fail like this.

Image

PR #2175, opened 2026-04-09
Implement client-session routing for Agent Web Socket Server

GeorgeNgMsft:
**Problem**
AgentWebSocketServer is a process-level singleton, but updateBrowserContext() was assigning its invoke handlers and connection/message callbacks as direct fields — overwriting them on every call. With multiple concurrent TypeAgent sessions, each new enable call would silently reroute all client traffic to the most recently activated session context, causing the wrong session to handle browser actions, web agent messages, and client connect/disconnect events.

**Solution**
Replace the shared callback fields with a Map<sessionId, SessionHandlers> so each session owns its own isolated handler set. Client connections are routed to the correct session by embedding sessionId in the WebSocket connection URL as a query parameter.

**Notes**
Added Architecture section documenting the WebSocket server, connection URL format, session routing, client type detection, and channel multiplexing


GeorgeNgMsft:
> This is a bigger change, we'll want to review this when working on the shell/browser extension clients

Actually, since the session operations aren't implemented on the clients yet, I think this is okay to merge in for now and iterate upon further after review.

PR #2178, opened 2026-04-09
Implement ClientIO interaction support on Agent-Server

GeorgeNgMsft:
**Async ClientIO Callbacks: Non-Blocking Deferred-Promise Pattern**
Three ClientIO methods — askYesNo, proposeAction, and popupQuestion — were previously implemented as blocking RPC invoke calls in the agent server. This meant the server-side command processing was tied to a specific client connection: a disconnect would permanently hang the operation, reconnecting clients couldn't recover pending questions, and popupQuestion was entirely disabled in server mode.

**This PR converts all three to a non-blocking deferred-promise pattern:**

The server generates a unique interactionId, logs the pending interaction to the DisplayLog, stores a deferred Promise in a new PendingInteractionManager, and sends a fire-and-forget requestInteraction notification to the client(s). Command processing suspends at await as before, but is no longer coupled to the connection.

The client responds via a new respondToInteraction dispatcher method, which resolves the deferred promise and resumes processing exactly where it left off.
On client disconnect, pending interactions owned by that connection are cancelled (rejecting their promises), and all callers treat this as a cancellation of the operation.
popupQuestion is now fully supported in server mode via broadcast to all connected clients, with first-responder resolution.

Pending interactions are logged to DisplayLog so they can be replayed to reconnecting or newly joining clients.
Existing callers (askYesNoWithContext, confirmTranslation, sessionContext.popupQuestion) required no changes — the ClientIO interface signatures are unchanged.


ts/packages/agentServer/server/src/sharedDispatcher.ts
```diff
@@ -229,8 +393,49 @@ export async function createSharedDispatcher(
...
             return dispatcher;
         },
+        respondToInteraction(response: PendingInteractionResponse): void {
+            debugInteraction(
+                `respondToInteraction: ${response.interactionId} type=${response.type}`,
+            );
+            const resolved = pendingInteractions.resolve(
+                response.interactionId,
+                response.value,
+            );
```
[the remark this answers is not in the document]
GeorgeNgMsft: The type system enforces the contract at compile time for all current callers, and adding runtime validation of internal RPC responses is defensive programming against a threat model that doesn't exist yet. If the system ever exposes respondToInteraction to untrusted clients, we will add it then.

ts/packages/dispatcher/dispatcher/src/dispatcher.ts
```diff
@@ -350,6 +353,16 @@ export function createDispatcherFromContext(
                 }
             });
         },
+        async respondToInteraction(
```
curtisman, line 356: I am wondering if this needs to be exposed in the API, can it all be hidden by the protocol?

ts/packages/agentServer/server/src/server.ts
```diff
@@ -139,6 +139,8 @@ async function main() {
                             connectionId: result.connectionId,
                             sessionId,
                             name: result.name,
+                            pendingInteractions:
```
curtisman, line 142: Doesn't the display log replay sufficient?

ts/packages/agentServer/server/src/sharedDispatcher.ts
```diff
@@ -121,18 +142,125 @@ export async function createSharedDispatcher(
...
+
+            // Broadcast to all connected clients
+            const notified = broadcast("requestInteraction", requestId, (cio) =>
+                cio.requestInteraction(request),
+            );
+            if (notified === 0) {
+                // No clients to handle it — resolve immediately without logging
+                // a pending entry (which would have no corresponding resolution).
+                return defaultValue ?? false;
```
curtisman, line 176: That would cause intermittent disconnection to miss some interactions?    Probably want to keep it pending until either user dismiss or the dispatcher session is closing down?

ts/packages/agentServer/server/src/sharedDispatcher.ts
```diff
@@ -121,18 +142,125 @@ export async function createSharedDispatcher(
...
+                throw new Error(
+                    "No connected clients available for proposeAction",
+                );
+            }
+
+            // Log only after we know at least one client was notified
+            context.displayLog.logPendingInteraction(request);
+            context.displayLog.saveQueued();
+
+            return pendingInteractions.create<unknown>(
```
curtisman, line 218: Instead of a timeout, at the server level, I think letting both the user or the requester (i.e. agent) able to explicit cancel will give them more control over the experience.

ts/packages/agentServer/server/src/sharedDispatcher.ts
```diff
@@ -121,18 +142,125 @@ export async function createSharedDispatcher(
...
+                cio.requestInteraction(request),
+            );
+            if (notified === 0) {
+                // No clients to handle it — resolve immediately without logging
+                // a pending entry (which would have no corresponding resolution).
+                return defaultValue ?? false;
+            }
+
+            // Log only after we know at least one client was notified
+            context.displayLog.logPendingInteraction(request);
```
curtisman, line 180: we don't log the message if it is not notified?  Wouldn't that be missed on replay?

ts/packages/agentServer/server/src/sharedDispatcher.ts
```diff
@@ -121,18 +142,125 @@ export async function createSharedDispatcher(
...
 
-        popupQuestion: async () => {
-            throw new Error("Not supported in server mode");
+        // ===== Async deferred pattern for blocking interactions =====
+        // Instead of blocking via callback(), we create a deferred promise
+        // and broadcast the request to all clients. The first client to
+        // respond via respondToInteraction resolves the promise.
+
+        askYesNo: async (requestId, message, defaultValue?) => {
+            const interactionId = randomUUID();
```
curtisman, line 152: if we log in up front, can we just use the sequence number?

ts/packages/dispatcher/dispatcher/src/dispatcher.ts
GeorgeNgMsft: Technically it could, if we moved respondToInteraction to use rpc.invoke (await) instead of rpc.send (fire and forget). However, the reason I chose this is because with the PendingInteractionManager and use of the FOF requestInteraction, we can actually extend functionality to allow "stashing" of interactions in the future.

As per our offline discussion, if the server asks the user a question, the next response can be interpreted by the LLM to determine if the user is responding to the question (and making a call to respondToInteraction) versus implicitly ignoring the question for now (sending a new command or following a different path of thought). These are things that happen naturally in a conversation such as coming back to a question later. 

Having this respondToInteraction() helps partially enable the scenario above, though it would take significantly more work to design and implement that interaction described above. I think it's a great future work item though

ts/packages/agentServer/server/src/server.ts
GeorgeNgMsft: Technically, you could retrieve the last question asked from the previous message in display log. However, that's adding additional (potentially fragile) extraction logic on an ordered stream of events. In addition, potential future capabilities like interruptions or moving operations into the background could make extracting from display log much more difficult.

I think it's cleaner to just store pending interactions in this array (which is also already filtered for the specific connection). No need to run any parsing logic.

ts/packages/agentServer/server/src/sharedDispatcher.ts
GeorgeNgMsft: That is a great callout, this scenario will come into play if the user sends a command, disconnects, and the server sends a question back. The user isn't currently able to reconnect and respond. 

In a scenario where the user was connected when the server sends the questions, but then disconnects, they will preserve the ability to respond until the timeout.

I will match the behavior in both scenarios so that the user can always respond until the timeout. I think setting a timeout is good policy for now, and something we can change later if we want to allow a question to hang indefinitely (Claude, I believe operates on a timeout as well).

ts/packages/agentServer/server/src/sharedDispatcher.ts
GeorgeNgMsft: The user can always cancel by simply responding, I think that part is fine. Though exposing an api to allow the agent to explicitly cancel (in other potential scenarios) is definitely useful. 

I prefer keeping a configurable timeout because it is good to guarantee completion of an interaction and resource cleanup. 

Both aren't mutually exclusive capabilities and we can tune them as we go.

ts/packages/agentServer/server/src/sharedDispatcher.ts
GeorgeNgMsft: This has been revaluated in the follow up PR along with not broadcasting on 0 users.

ts/packages/agentServer/server/src/sharedDispatcher.ts
GeorgeNgMsft: I think this goes back to the conversation about having PendingInteractionRequests versus parsing the logs. I think maintaining a queue of PendingInteractionRequests is better structure that allows for more extensibility than log extraction (especially for the long run).

PR #2180, opened 2026-04-10
Fix multiple bugs: XSS vulnerabilities, operator precedence, event listener loss, race condition

hobostay:
## Summary

- **Fix XSS in `browserViewManager.ts`**: Three locations directly interpolated URLs and error descriptions into `document.body.innerHTML` via `executeJavaScript` without escaping. Malicious URLs containing `<script>` tags or HTML entities could execute arbitrary JavaScript in the Electron renderer. Fixed by HTML-escaping all interpolated values before use.

- **Fix operator precedence bug in `videoActionHandler.ts`**: The expression `"text" + statusData.failure_reason ?? ""` evaluates as `("text" + statusData.failure_reason) ?? ""` because `+` has higher precedence than `??`. This makes the nullish coalescing operator dead code — when `failure_reason` is `undefined`, the string `"undefined"` is displayed instead of an empty string. Added parentheses: `(statusData.failure_reason ?? "")`.

- **Fix unhandled promise rejection in `setContent.ts`**: The `Promise.all(promises).then(...)` for loading CSS into iframes has no `.catch()` handler. If any CSS fetch fails, the iframe never gets `srcdoc` content at all. Added a `.catch()` that renders the message content even without CSS.

- **Fix event listener loss in `setContent.ts`**: `contentElm.innerHTML += contentHtml` serializes and re-parses the entire DOM, destroying all existing event listeners on child elements (including previously-attached link click handlers). Replaced with `insertAdjacentHTML("beforeend")` which preserves existing DOM nodes.

- **Fix race condition in `webSocketAPI.ts`**: The `onclose` handler unconditionally calls `createWebSocket()` when `autoReconnect` is true. If the socket closes rapidly multiple times (e.g., network flapping), multiple concurrent reconnection attempts are spawned, potentially creating duplicate WebSocket connections. Added a `reconnecting` flag to serialize reconnection attempts.

## Test plan

- [ ] Verify browser tabs still show error pages correctly when URLs fail to load
- [ ] Verify video agent displays failure reason correctly when `failure_reason` is undefined vs set
- [ ] Verify iframe content renders even when CSS fetch fails
- [ ] Verify link click handlers still work after multiple messages are appended in chat
- [ ] Verify WebSocket reconnects properly after connection loss without creating duplicates

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ts/packages/shell/src/main/browserViewManager.ts
```diff
@@ -196,8 +196,20 @@ export class BrowserViewManager {
                 // only show the error if it's for the page the user was asking
                 // it's possible some other resource failed to load (image, script, etc.)
                 if (validatedURL === options.url) {
+                    const safeUrl = options.url
```
robgruen: Since this is pattern is repeated for safeErr, can we make this a utility method somewhere and call it.

robgruen: @copilot , can you address this comment

robgruen: did it myself

PR #2201, opened 2026-04-16
Add ToggleMouseSonar and AdjustVolume desktop actions

TalZaccai:
This branch (based on `talzacc/action_json_schema`) demonstrates how easy it is to add new actions after the architecture refactoring in #2196. Two new actions are added, showcasing the two main patterns described in the [README](dotnet/autoShell/README.md#adding-a-new-action).

### ToggleMouseSonar — Option A: Registry Toggle

Enables/disables the Windows "Find my pointer" ring that appears when pressing Ctrl.

**Total lines added: 1** (in `MouseSettingsHandler.cs`)

```csharp
AddRegistryToggleAction("ToggleMouseSonar", new RegistryToggleConfig(
    @"Control Panel\Mouse", "MouseSonar", "enable",
    OnValue: "1", OffValue: "0", ValueKind: RegistryValueKind.String, DisplayName: "Mouse Sonar"));
```

No custom handler logic. The `AddRegistryToggleAction` helper does everything — reads the `enable` parameter, writes the registry value, returns an `ActionResult`.

### AdjustVolume — Option B: Typed Action with Custom Logic

Adjusts volume relatively ("turn it up a bit") instead of setting an absolute value.

**Steps:**
1. Define TS schema type in `actionsSchema.ts`
2. Add handler in `AudioActionHandler.cs` — reads current volume, applies delta, clamps 0–100, saves for restore

```typescript
export type AdjustVolumeAction = {
    actionName: "AdjustVolume";
    parameters: {
        direction: "up" | "down";
        amount?: number; // default 10
    };
};
```

The C# `AdjustVolumeParams` record is **automatically generated** by the Roslyn source generator from the `.pas.json` schema — no manual C# type definition needed.

### Test Coverage

- 6 new tests for `AdjustVolume` (up/down, default amount, clamping, volume save)
- 2 new tests for `ToggleMouseSonar` (enable/disable registry writes)

### DIFF
--- a/dotnet/autoShell.Tests/AudioActionHandlerTests.cs
+++ b/dotnet/autoShell.Tests/AudioActionHandlerTests.cs
@@ -135,6 +135,80 @@ public class AudioActionHandlerTests
         _audioMock.Verify(a => a.SetMute(false), Times.Once);
     }
 
+    // --- AdjustVolume ---
+
+    /// <summary>
+    /// Verifies that AdjustVolume "up" increases volume by the specified amount.
+    /// </summary>
+    [Fact]
+    public void AdjustVolume_Up_IncreasesVolume()
+    {
+        _audioMock.Setup(a => a.GetVolume()).Returns(50);
+
+        var result = _handler.Handle("AdjustVolume", JsonDocument.Parse("""{"direction":"up","amount":15}""").RootElement);
+
+        Assert.True(result.Success);
+        _audioMock.Verify(a => a.SetVolume(65), Times.Once);
+    }
+
+    /// <summary>
+    /// Verifies that AdjustVolume "down" decreases volume by the specified amount.
+    /// </summary>
+    [Fact]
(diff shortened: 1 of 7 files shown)


robgruen:
for the registry toggle action, once you set the value does it immediately take effect or do we need to broadcast some sort of WM_SETTINGCHANGED for it to work first?

robgruen:
BTW, 😍

TalZaccai:
> for the registry toggle action, once you set the value does it immediately take effect or do we need to broadcast some sort of WM_SETTINGCHANGED for it to work first?

It does not take affect immediately, but this might be a wider fix than just this specific toggle, I'll look into fixing it in a separate PR.

robgruen:
Might be this: WM_SETTINGCHANGE message (Winuser.h) - Win32 apps | Microsoft Learn<https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-settingchange>

PR #2205, opened 2026-04-16
Fix multiple bugs: XSS, operator precedence, event listener loss, rac…

robgruen:
…e condition

- Fix XSS in browserViewManager.ts: HTML-escape URLs and error messages before interpolating into innerHTML via executeJavaScript (3 locations)
- Fix operator precedence bug in videoActionHandler.ts: `+` has higher precedence than `??`, making `failure_reason ?? ""` dead code. Added parentheses to correct the logic
- Fix unhandled promise rejection in setContent.ts: added .catch() handler for CSS loading in iframe so content is still shown on fetch failure
- Fix event listener loss in setContent.ts: replaced `innerHTML +=` with `insertAdjacentHTML("beforeend")` to avoid destroying existing DOM nodes and their event listeners
- Fix race condition in webSocketAPI.ts: added `reconnecting` flag to prevent multiple concurrent reconnection attempts on rapid socket closes

### DIFF
--- a/ts/packages/agents/video/src/videoActionHandler.ts
+++ b/ts/packages/agents/video/src/videoActionHandler.ts
@@ -198,7 +198,7 @@ function createVideoPlaceHolder(
                     container.innerText = "❌ Failed to retrieve video content.";
                 }
             } else {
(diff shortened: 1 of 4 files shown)


robgruen:
@copilot resolve the merge conflicts in this pull request

PR #2210, opened 2026-04-16
Fix TODOs, optimize functions, and enhance test coverage

robgruen:
[Copilot is generating an outline...]


robgruen:
separated into multiple smaller prs

PR #2277, opened 2026-04-28
Shell IntelliSense & demo polish fixes

TalZaccai:
## Shell IntelliSense

- Per-prefix completion mode: inline ghost for free text, dropdown for `@`-commands (regardless of user's inline preference).
- **Enter** accepts the highlighted dropdown item (in addition to **Tab**); still submits when no menu is open.
- **Enter** with an inline ghost suggestion submits the typed text only (does not silently accept the suggestion).
- Trailing space appended after accepted completions so the next level fetches automatically.
- Leading space inserted when accepting at a non-separator boundary so `@shell` + `break` yields `@shell break ` not `@shellbreak `.
- **Ctrl+Space** reopens the menu after Escape (or from cold).
- New `CompletionController.show()` API to force a fresh fetch bypassing the dedup guard.
- Inline ghost-text suggestion is stripped from the message before sending.

## Search menu fixes

- `RemoteSearchMenuUI` mirrors `selected` / `itemCount` / `firstUpdate` locally so `selectCompletion()` returns the highlighted item synchronously (fixes Enter accepting nothing in shell mode).
- Mouse wheel scrolls the dropdown without snap-to-selected; new `searchMenuScroll` IPC plumbs wheel events to the Electron-hosted menu (`localSearchMenuUI.scrollBy`, `electronSearchMenuUI`, preload, view host).

## Subcommand-menu visibility fix

- Dispatcher: subcommand groups returned by the table-fallback branch now use `separatorMode: "optionalSpace"` (was conditionally `"space"`), matching the descriptor branch. Fixes silent menu-hide after accepting agents like `@shell` / `@list` / `@browser` whose top-level command table has no `defaultSubCommand`.

## Conversation commands

- New `@conversation` command (`new | list | info | switch | rename | delete | next | prev`) — discoverable via the `@` menu, dispatches the same `manage-conversation` client action the NL handler uses.
- `switch` with no name and the new `next` / `prev` actions cycle through conversations; chat input shows a "Switching conversation…" busy state while loading.
- Single-conversation (self-hosted) mode: `prev` / `next` / unnamed `switch` show an inline yellow warning instead of silently no-op'ing.
- Schema: added `NextConversationAction` and `PrevConversationAction`; strengthened `ListConversationAction` / `SwitchConversationAction` descriptions for better LLM routing.
- Removed renderer-side `conversationCommands.ts` interceptor — the dispatcher is now the single source of truth (the renderer only normalises the legacy `/conversation` slash alias to `@conversation`). Eliminates the recurring drift between renderer and dispatcher implementations of the same command.

## Demo Mode

- **Esc** breaks a running demo (handler installed on main + chat web contents while active).
- New `@shell break` command does the same.
- `[Demo Mode]` pill in the title bar while a demo is running.
- Chat input placeholder reflects demo state: "Demo running… Esc to break" / "Ctrl+Right to continue • Esc to break demo".
- Demo file is opened *before* setting `demoRunning = true`, so cancelling the picker no longer locks out future runs.
- Concurrent `@shell run` invocations show an inline yellow warning instead of interleaving lines (sync `demoStarting` flag closes the async startup race).

## Docs

- New `docs/content/help/commandReference.md` and `docs/content/help/shellShortcuts.md` covering the commands and shortcuts touched by this PR.


robgruen:
Can we add the new commands to https://github.com/microsoft/TypeAgent/blob/main/docs/content/help/commandReference.md with new commands

ts/packages/dispatcher/dispatcher/src/context/system/handlers/conversationCommandHandlers.ts
```diff
@@ -0,0 +1,151 @@
...
+    public async run(context: ActionContext<CommandHandlerContext>) {
+        dispatchManageConversation(context, { subcommand: "info" });
+    }
+}
+
+class ConversationSwitchCommandHandler implements CommandHandler {
+    public readonly description = "Switch to an existing conversation";
+    public readonly parameters = {
+        args: {
+            name: {
```
robgruen, line 73: Thoughts: could make this optional and if you don't supply it, you go back to the previous conversation (or the next one in the list)?  Or add prev/next conversation command?

ts/packages/dispatcher/dispatcher/src/context/system/handlers/conversationCommandHandlers.ts
```diff
@@ -0,0 +1,151 @@
...
+    CommandHandlerContext,
+    getRequestId,
+} from "../../commandHandlerContext.js";
+
+// Each handler dispatches the same "manage-conversation" client action
+// emitted by the natural-language conversationActionHandler — so the CLI
+// and Shell already know how to render the result.
+function dispatchManageConversation(
+    context: ActionContext<CommandHandlerContext>,
+    payload: { subcommand: string; name?: string; newName?: string },
```
robgruen: nit: make this it's own type/interface?

ts/packages/dispatcher/dispatcher/src/helpers/completion/controller.ts
```diff
@@ -41,6 +42,7 @@ export interface CompletionController {
     update(input: string, direction?: CompletionDirection): void;
     accept(): void;
     dismiss(input: string, direction?: CompletionDirection): void;
+    show(input: string, direction?: CompletionDirection): void;
```
robgruen, line 45: Good call adding this!

ts/packages/dispatcher/dispatcher/src/helpers/completion/session.ts
```diff
@@ -278,6 +278,23 @@ export class PartialCompletionSession implements CompletionController {
         this.setCompletionState(undefined);
     }
 
+    /**
+     * Explicitly reactivate the completion menu after a dismiss
+     * (Ctrl+Space).  Clears dismissAnchor (so a refetch with the same
```
robgruen, line 283: Maybe we should document somewhere all of the keyboard shortcuts so they are discoverable. Maybe in the readme or in a doc here: https://github.com/microsoft/TypeAgent/tree/main/docs/content/help

ts/packages/shell/src/main/demo.ts
```diff
@@ -22,64 +22,148 @@ async function openDemoFile(window: BrowserWindow) {
...
+        () => {
+            if (ipcCallback)
+                ipcMain.removeListener("send-input-text-complete", ipcCallback);
+            abort.unregister();
+        },
+    );
+}
+
+let demoRunning = false;
+let demoAborted = false;
```
robgruen: couldn't we just use an instance of DemoState to reflect all of these?  (need to add one more DemoState enum:  aborted).

ts/packages/shell/src/preload/electronTypes.ts
```diff
@@ -106,6 +106,7 @@ export interface Client {
     systemNotification?(message: string, id: string, timestamp: number): void;
     conversationChanged?(conversationId: string, name: string): void;
     markHistoryEntries?(): void;
+    demoStateChanged?(state: "running" | "paused" | "idle"): void;
```
robgruen: DemoState type

ts/packages/shell/src/renderer/src/chat/chatView.ts
```diff
@@ -1026,6 +1026,19 @@ export class ChatView {
         this._voiceBanner.textContent = active ? "Claude Focus" : "Voice Mode";
     }
 
+    public setDemoState(state: "running" | "paused" | "idle"): void {
```
robgruen: DemoState type

ts/packages/shell/src/renderer/src/partial.ts
```diff
@@ -184,14 +197,22 @@ export class PartialCompletion {
...
+        this.searchMenu.switchMode(liveInline);
         // Always full render after mode switch (new UI instance).
         const state = this.controller.getCompletionState();
         if (state) {
             this.searchMenu.render(state.prefix, state.items);
         }
     }
 
+    private getEffectiveInline(): boolean {
+        return this.previousInput.startsWith("@") ? false : this.userInline;
```
robgruen: since we may switch @ to / let's make this a constant somewhere and reuse it here.

ts/packages/shell/src/renderer/src/partial.ts
```diff
@@ -334,12 +355,37 @@ export class PartialCompletion {
...
+        const lastChar = quotedText.slice(-1);
+        const appendSpace = /[A-Za-z0-9_"'\)\]]/.test(lastChar);
+
+        // Prepend a leading space when the menu's prefix is empty and the
+        // character immediately preceding the insertion point is
+        // alphanumeric (e.g. accepting a subcommand right after "@shell"
+        // with no separator yet typed).  Otherwise the inserted text
+        // would fuse with the previous token (e.g. "@shellbreak ").
         const offset = this.getCurrentInput().length - completionPrefix.length;
+        const charBefore =
```
robgruen, line 396: Let's put a comment somewhere here that we need to add support for RTL languages

robgruen (approved):

PR #2291, opened 2026-05-04
TypeAgent chat panel for VS Code

TalZaccai:
# TypeAgent chat panel for VS Code

Brings the TypeAgent chat experience into VS Code as a first-class host
alongside the Electron shell. Same dispatcher, same agents, same chat UI ΓÇö
now embedded in VS Code with multi-panel session sharing, a scripted demo
runner, and a polished reconnect/cancel UX.

**Branch:** `talzacc/vscode-shell-chat` ┬╖ 85 files ┬╖ ~14.2k / -650 vs `main`.

## VS Code chat experience
- [x] Sidebar + side-panel chats; multiple panels share one session.
- [x] Sessions persist across VS Code reload; full history replays on rejoin.
- [x] Connection ribbon shows live status (Connected / Disconnected / Demo Mode).
- [x] Chord keybindings: new chat, clear chat, continue demo, cancel demo.
- [x] Theme-aware (follows the active VS Code theme).

## Chat UX polish (shared with Electron shell)
- [x] **Stop button** actually aborts the in-flight request and shows `ΓÜá Cancelled`.
- [x] **Reconnect**: single in-place countdown (`Disconnected ΓÇö retrying in Ns ┬╖ <error>`) instead of error spam in the chat area.
- [x] **IntelliSense in chat-ui webview**: ports the shell's partial-completion engine into the shared `chat-ui` package so the new vscode-shell host gets inline ghost + dropdown, arrow-key cycling, Ctrl+Space re-summon, Tab-snaps-to-first on auto-opened dropdowns, and Enter-submits-typed-text in inline mode.
- [x] **Multi-panel session sharing**: when more than one panel is joined to the same conversation, all of them stay in sync ΓÇö peer panels show bubbles, status updates, final metrics, and history-replay state identical to the originator's.
- [x] **Visual polish**: shell-style metrics row, JSON syntax highlighting, hover-only timing tooltips, brighter agent-bubble + ANSI contrast, real username, source-keyed avatars driven end-to-end by each agent's manifest `emojiChar` (forwarded through the agent-server bridge ΓåÆ webview ΓåÆ chat-ui).

## Demo runner (vscode-shell)
- [x] Run a `.txt` script from any text editor right-click ΓåÆ "Run TypeAgent Demo Script".
- [x] Lines type themselves character-by-character into the input.
- [x] `@pauseForInput` markers stop on a paused state; Ctrl+ΓåÆ continues.
- [x] Esc cancels at any time ΓÇö paused, between lines, or mid-typing.
- [x] Status ribbon reflects `[Demo Mode (Running)/(Paused)]`; pause hint shown as input ghost.
- [x] No leaked timers, no message corruption from racing key handlers.
- [x] Sync `demoStarting` flag closes a startup race when two invocations fire close together during the file-picker await.

(Independent of the shell-side `@shell run` / `@shell break` demo improvements in PR #2277 ΓÇö different host, different code path.)

## VS Code integrations (new agent actions)
- [x] Open / create files (`newFile` writes to disk before opening so the file persists).
- [x] Switch to the editor, change theme, open settings, etc., via natural language.
- [x] Conversation management (new / rename / switch) exposed via a new `code-vscode-shell` sub-schema in the `code` agent. **Opt-in**: `defaultEnabled: false`; new vscode-shell users run `@config schema code-vscode-shell on` once (documented in `ts/packages/vscode-shell/README.md`).

## New `github-cli` agent
- [x] Talk to the `gh` CLI in natural language (`@github-cli list my prs`, ΓÇª).
- [x] Schema generated using the new onboarding CLI-help-crawl workflow ΓÇö same approach can be reused for any other CLI tool.

## Package architecture
- [x] **`@typeagent/completion-ui`** is now a zero-runtime-dependency shared widget package. The canonical `SearchMenuItem` type lives here; `agent-dispatcher` imports (and re-exports) it for backwards compatibility. Lightweight downstream consumers (Chrome extension, future hosts) no longer drag the entire dispatcher graph through the menu.

## Server / dispatcher infrastructure
- [x] **Per-conversation `SharedDispatcher`** in `agentServer` that multiplexes ClientIO across all clients joined to the same conversation; also bridges `processCommand`'s RPC return value into a server-emitted `commandComplete` notify so peer panels stay in sync with the originator.
- [x] **`DisplayLog`** captures display events so a freshly-joined client can replay the conversation.
- [x] Idle conversations evicted by a new `ConversationManager`.
- [x] Reasoning engine made configurable (Copilot / Claude).
- [x] WebSocket server no longer crashes when sending to a closed socket.
- [x] **Schema cleanup per code-review feedback**: split the `code` agent's `NewFileAction` into `NewCodeFileAction` / `NewMarkdownFileAction` / `NewTextFileAction` so cross-agent `newFile` collisions don't need defensive dispatcher patches; stripped negative-assertion `DO NOT USE` walls from action descriptions in favor of positive `User:` / `Agent:` examples.

## Bug fixes worth calling out
- [x] **Cross-session callId collision** in the `code` agent ΓÇö `nextCallId` was per-session but the pending-calls map is module-global, so concurrent sessions collided and routed responses to the wrong place. Fixed by making the counter module-global.
- [x] Stop button previously sent the webview's clientRequestId where the dispatcher expected its own server-side requestId, so cancel did nothing. Now translated correctly.

## Verification
- [x] B4 smoke-test pass across all flows above (single + multi-tab, ghost text, completion cycling, session restore, history replay, demo state, stop button).
- [x] 3 parallel code-review agent passes; findings actioned (one false positive documented).
- [x] Prettier across all changed files.

## Try it
1. Start the server: `cd ts/packages/agentServer/server && npm start` (`ws://localhost:8999`, ~3 min full init).
2. Install the extension: `cd ts/packages/vscode-shell && npm run deploy:local`.
3. Click the TypeAgent activity-bar icon ΓÇö ribbon should show Connected.
4. Open a second panel (Ctrl+K Ctrl+T) ΓÇö sessions share live.
5. Stop the server ΓåÆ single in-place countdown ribbon. Restart ΓåÆ clears.
6. Right-click `ts/packages/shell/demo/vscode_shell_extension_part1.txt` ΓåÆ Run TypeAgent Demo Script. Ctrl+ΓåÆ at pauses; Esc to cancel.


ts/packages/coda/src/handleVSCodeActions.ts
```diff
@@ -268,12 +269,64 @@ export async function handleBaseEditorActions(
...
+            const language: string | undefined = actionData.language;
+
+            const workspaceFolder = vscode.workspace.workspaceFolders?.[0];
+            const hasRealName =
+                typeof fileName === "string" &&
+                fileName.trim().length > 0 &&
+                fileName.trim().toLowerCase() !== "untitled";
+
+            if (workspaceFolder && hasRealName) {
+                const fileUri = vscode.Uri.joinPath(
```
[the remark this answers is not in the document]
TalZaccai: Tracking this in a follow-up. The fix is non-trivial — the action schema has no folder parameter, so we need either (a) infer the workspace folder from the active editor, or (b) add a folder param and route accordingly. Out of scope for this PR (pre-existing behavior in the coda agent, not introduced here).

dotnet/autoShell/Services/WindowsWindowService.cs
```diff
@@ -48,6 +48,23 @@ private struct RECT
     [DllImport(NativeDlls.User32)]
     private static extern bool ShowWindow(IntPtr hWnd, uint nCmdShow);
 
+    [DllImport(NativeDlls.User32)]
+    private static extern bool IsIconic(IntPtr hWnd);
```
robgruen: nit: sort these alphabetically?

ts/packages/agents/code/src/vscode/editorCodeActionsSchema.ts
```diff
@@ -13,7 +13,9 @@ export type EditorCodeActions =
     | EditorActionSaveCurrentFile
     | EditorActionSaveAllFiles;
 
-// Action to create a new file in the editor
+// DEPRECATED in favor of the main code agent's `newFile` action. Do NOT use this for new
```
robgruen: Deprecate like so:
/**
 * @deprecated Use `NewConfig` instead.
 */
type OldConfig = {
  foo: string;
};

ts/packages/agents/code/src/vscode/vscodeShellActionsSchema.ts
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export type VscodeShellActions =
```
robgruen: Naming might be confusing...consider **VSCodeTypeAgentConversationActions** or something similar.  VS Code has it's own "shell" so need some sort of disambiguation.

ts/packages/agents/code/src/vscode/vscodeConversationActionsSchema.ts
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export type VscodeShellActions =
+    | NewConversationAction
```
robgruen, line 5: For later: to have parity with other agentic systems we will want "Fork", "Rewind", etc. conversation.

ts/packages/agents/code/src/codeActionHandler.ts
```diff
@@ -57,28 +80,29 @@ async function updateCodeContext(
...
             return;
         }
+        if (agentContext.enabled.size === 0) {
+            sharedWebSocketRefCount++;
+        }
+        agentContext.enabled.add(schemaName);
 
-        if (!context.agentContext.webSocketServer) {
+        if (!sharedWebSocketServer) {
             const port = parseInt(process.env["CODE_WEBSOCKET_PORT"] || "8082");
```
robgruen, line 95: Let's put a TODO in here that we need to NOT use a hardcoded port and should instead get a port to use from the dispatcher.

ts/packages/agents/code/src/codeActionHandler.ts
```diff
@@ -267,6 +297,32 @@ async function executeCodeAction(
         return undefined;
     }
 
+    // TypeAgent Shell conversation-management actions are routed back to
+    // the originating extension webview via takeAction (which targets the
+    // requesting client only). They do NOT use the Coda WebSocket bridge.
+    if (
+        action.actionName === "newConversation" ||
```
robgruen: instead of the strings here let's use NewConversationAction .actionName, etc.

ts/packages/agents/code/src/codeActionsSchema.ts
```diff
@@ -86,17 +95,33 @@ export type CodeLanguage =
...
+//   "new markdown file called notes.md"
+//     → { actionName: "newFile", parameters: { fileName: "notes.md", language: "markdown", content: "" } }
+//
+// DO NOT route file-creation requests to dispatcher.reasoning, dispatcher.unknown,
+// dispatcher.clarify, or any code-editor/code-general sub-schema. Always pick newFile here.
 export type NewFileAction = {
     actionName: "newFile";
     parameters: {
-        fileName: string; // If not filename is provided, "Untitled" is used
+        fileName: string; // If no filename is provided, "Untitled" is used
```
robgruen: a better way to declare this is like so:

// The file name or "untitled"
fileName: string | "untitled";

ts/packages/agents/code/src/codeActionsSchema.ts
```diff
@@ -86,17 +95,33 @@ export type CodeLanguage =
...
         language: CodeLanguage; // plaintext, html, python, javaScript, typeScript, markdown etc.
```
robgruen: since the type is "CodeLanguage", you don't need the superfluous comment at the end of the line (typically).

ts/packages/agents/code/src/codeManifest.json
```diff
@@ -30,15 +30,15 @@
     },
     "code-general": {
       "schema": {
-        "description": "Code agent that helps you perform actions like finding a file, go to a symbol or a line in a file, Open new vscode window, show the command palette, user settings json etc.",
+        "description": "Code agent that helps you navigate within an already-open editor: find a file in the workspace, go to a symbol or line, open a new vscode window, show the command palette, open user settings json, etc. Does NOT create files (use the main 'code' schema's newFile action) and does NOT handle TypeAgent chat conversations.",
```
robgruen: Try to avoid these negative assertions (and if you do have to use them, only use them as negative examples on action definitions and not on the agent description).  The other actions may come or go so having these here can be distracting for the LLM...if there's an action that's "stealing" attention then we should tune the action schema of both the thief and the victim.

ts/packages/agents/code/src/codeActionsSchema.ts
robgruen: alternatively, wouldn't a file extension be more appropriate here?  cause if it's *.py you know what it is (and same for all of the other file extensions).  I think that would be much easier for the LLM to grok.

ts/packages/agents/code/src/codeActionsSchema.ts
```diff
@@ -86,17 +95,33 @@ export type CodeLanguage =
...
 
-// Create a new file, this is not same as opening a file or finding a file in the workspace
+// ACTION: Create a new file in the editor.
+// USE THIS for any "create a new <lang> file called X with <content>", "make a new file",
+// "new typescript/python/markdown file" — for both untitled scratch files and named files.
+// This is the SINGLE canonical action for creating files. Do NOT emit a schema name like
+// "code-editor" or "code-general" as the actionName — always use "newFile" exactly.
+// Distinct from opening or finding an existing file in the workspace.
+//
+// EXAMPLES (always route these to newFile):
```
robgruen: Having a grammar for these examples will prevent you from having to list these examples.  Also, it's more than likely that another agent will have a "NewFileAction" (i.e. word).  A better approach, which you should test, is to offer multiple actions that cover everything you want and that the LLM wants to translate to.  For example I would get rid of this and replace it with:   NewCodeFileAction, NewMarkdownFileAction. NewTextFileAction.  See my schema best practices list: ts/packages/agents/onboarding/src/schemaGen/schemaGenHandler.ts#L26-L63

ts/packages/agents/code/src/codeManifest.json
```diff
@@ -59,6 +59,15 @@
         "schemaFile": "../dist/extensionSchema.pas.json",
         "schemaType": "CodeWorkbenchExtensionActions"
       }
+    },
+    "code-vscode-shell": {
+      "defaultEnabled": false,
+      "schema": {
+        "description": "Manage TypeAgent Shell conversations (chat tabs in the TypeAgent VS Code Shell extension): create a new conversation, rename the current conversation, or switch to an existing conversation by name. Trigger keywords: 'conversation', 'chat', 'chat tab', 'new conversation', 'rename conversation', 'switch conversation'. Does NOT handle files, windows, browser tabs, editor tabs, programs, or onboarding.",
```
robgruen: This smells very mischievous, does it work without the "trigger keywords"?  Maybe the "code-vscode-shell" name isn't correct...maybe it should be code-vscode-conversations, or something like that?

ts/packages/agents/desktop/src/actionsSchema.ts
```diff
@@ -75,7 +75,12 @@ export type MinimizeWindowAction = {
     };
 };
 
-// Sets focus to a program window on a Windows Desktop
+// Sets focus to a program window on a Windows Desktop.
```
robgruen, line 78: again this feels overly attention seeking and best to leave out.  Instead of doing it like this I would just give it an example:

// Example:
// User: switch to <applicationName>
// Agent: { actionName: "SwitchTo", parameters: {  } }

ts/packages/agents/desktop/src/actionsSchema.ts
```diff
@@ -135,11 +140,18 @@ export type ChangeThemeModeAction = {
...
+// USE THIS only when the user explicitly references the WINDOWS DESKTOP, WALLPAPER, SYSTEM, or
+// PERSONALIZATION theme, e.g. "change my windows theme to Glow", "apply the Sunrise wallpaper theme".
+// DO NOT USE for VS Code / editor / IDE color themes (e.g. "Monokai", "Solarized Dark", "Dark+",
+// "Light+") — those are editor color themes and must be handled by the code agent's
+// `changeColorScheme` action, NOT by ApplyTheme.
 export type ApplyThemeAction = {
     actionName: "ApplyTheme";
     parameters: {
-        filePath: string; // The theme name or .theme file path to apply (use "previous" to revert)
+        filePath: string; // The Windows theme name or .theme file path to apply (use "previous" to revert). Never a VS Code editor theme like "Monokai".
```
robgruen: Here what you can do (if the LLM  really *WANTS* to put a theme name in, just add that as an optional parameter?: string and then just never use it.  That way you can remove the "never comment statement"

ts/packages/agents/player/src/agent/playerSchema.ts
```diff
@@ -54,7 +54,11 @@ export interface PlayRandomAction {
     };
 }
 
-// Play a specific track
+// Play a specific track. ONLY for music playback requests where the user
+// clearly names a song (e.g. "play Bohemian Rhapsody", "play Yesterday by
+// the Beatles"). DO NOT use for "switch to <app>", "focus on <app>",
```
robgruen: We should probably leave it the way it was.  *maybe* add the example but even that's touchy.  Definitely don't want action pollution from other agent's actions in here, that will cause spaghetti references in the long run and distract attention heads.

ts/packages/agentServer/server/src/conversationManager.ts
```diff
@@ -88,24 +88,54 @@ export async function createConversationManager(
     idleTimeoutMs: number = DEFAULT_IDLE_TIMEOUT_MS,
 ): Promise<ConversationManager> {
     const conversationsDir = path.join(baseDir, CONVERSATIONS_DIR);
-    await fs.promises.mkdir(conversationsDir, { recursive: true });
 
-    // Migrate old on-disk layout: "server-sessions/" → "conversations/"
+    // Migrate old on-disk layout: "server-sessions/" → "conversations/".
```
robgruen, line 95: Let's add a note to deprecate this migration code at some point in the future.

ts/packages/chat-ui/src/chatPanel.ts
```diff
@@ -9,13 +9,73 @@
...
+    type PcPost,
+} from "./partialCompletion.js";
+
+/**
+ * Default per-agent emoji map used when a host calls add/replaceAgentMessage
+ * without an explicit `sourceIcon`. Sourced from the manifest emojiChar values
+ * in ts/packages/agents/* /src/*Manifest.json. Hosts can extend or override
+ * via `ChatPanel.setAvatarMap`.
+ */
+export const DEFAULT_AVATAR_MAP: Readonly<Record<string, string>> = {
```
robgruen, line 46: What's the plan for keeping this in sync?

The Shell has a method that can get the agent icons from the dispatcher at run time.  We should replicate that functionality here.

ts/packages/chat-ui/src/chatPanel.ts
```diff
@@ -1236,39 +2402,87 @@ class AgentMessageContainer {
...
+        if (label) {
+            this.nameSpan.textContent = label;
+            this.actionDerivedName = label;
+        }
         this.nameSpan.classList.add("clickable");
-        this.nameSpan.title = "Click to show action JSON";
+        this.nameSpan.title = "Click to show / hide action JSON";
     }
 
     private toggleActionData() {
```
robgruen, line 2475: This chatPanel.ts file is MASSIVE.  Is there any way to logically break it into smaller chunks (i.e. agent message stuff vs. user messages vs. styling, etc.)?

ts/packages/dispatcher/dispatcher/src/reasoning/claude.ts
```diff
@@ -328,11 +328,27 @@ function getClaudeOptions(
                 if (
                     validationResult.message.startsWith("Unknown action name:")
                 ) {
+                    // Cross-schema hint: search other active schemas for
```
robgruen: I'd like to understand this a bit more. I don't know if we should add this.  "UnknownAction" should only ever be registered by the dispatcher, it's a *reserved* action.  Finding actions with the same name in other active schemas circumvents all of the logic built into the dispatcher and I don't know if we want to enable that.  We should think very carefully about that.  Was this code added because of the NewFileAction being stolen by other schemas?

ts/packages/dispatcher/dispatcher/src/reasoning/copilot.ts
```diff
@@ -410,7 +410,34 @@ function getCopilotSessionConfig(
 
             const validationResult = validator.validate(actionJson);
             if (!validationResult.success) {
-                throw new Error(validationResult.message);
+                // Cross-schema hint: if the (unknown) actionName is
```
robgruen: See my previous comment in the claude reasoning.

ts/packages/dispatcher/dispatcher/src/translation/actionSchemaJsonTranslator.ts
```diff
@@ -109,9 +109,35 @@ export function createActionSchemaJsonValidator<T extends TranslatedAction>(
                 if (value.actionName === undefined) {
                     return error("Missing actionName property");
                 }
-                const actionSchema = actionSchemaGroup.actionSchemas.get(
+                let actionSchema = actionSchemaGroup.actionSchemas.get(
                     value.actionName,
                 );
+                if (
```
robgruen: This is another piece of logic I would like to understand.  When everything works correctly it seems like an unnecessary patch.  If there are hallucinations with schema names we should understand why those happened before we put in hallucination guards within the dispatcher.  We do support this type of namespace lookup for entity resolution which is a more narrow field correction.  I don't know if it applies to actions, let's discuss.

ts/packages/dispatcher/dispatcher/src/translation/actionSchemaJsonTranslator.ts
```diff
@@ -124,7 +150,12 @@ export function createActionSchemaJsonValidator<T extends TranslatedAction>(
                         value.schemaName = injectedSchemaName;
                         return success(value);
                     }
-                    return error(`Unknown action name: ${value.actionName}`);
+                    const validNames = Array.from(
+                        actionSchemaGroup.actionSchemas.keys(),
+                    ).slice(0, 30);
+                    return error(
+                        `Unknown action name: "${value.actionName}". The actionName must be one of the action types defined in the schema (e.g. "newFile", "createFile"), NOT a schema/agent name (e.g. "code", "code-editor"). Available actions in this schema include: ${validNames.join(", ")}.`,
```
robgruen: Shouldn't be polluting specific agent/action names here.

ts/packages/vscode-shell/src/webview/globals.d.ts
robgruen, line 1: should this be checked in?

ts/packages/vscode-shell/src/agentServerBridge.ts
robgruen, line 1: can we break this file up?

dotnet/autoShell/Services/WindowsWindowService.cs
TalZaccai: Done — sorted alphabetically (consts and the EnumWindowsProc delegate kept at the top).

ts/packages/agents/code/src/codeActionHandler.ts
TalZaccai: Done.

ts/packages/agents/code/src/codeActionsSchema.ts
TalZaccai: Done.

ts/packages/agents/code/src/codeActionsSchema.ts
TalZaccai: Done.

ts/packages/agents/code/src/codeManifest.json
TalZaccai: Done — stripped from `code-general` and `code-editor` descriptions. (`code-vscode-shell` description still has trigger-keyword/disambiguation language; will address as part of your separate comment on that entry.)

ts/packages/agents/desktop/src/actionsSchema.ts
TalZaccai: Done — added `themeName?: string`.

ts/packages/agents/player/src/agent/playerSchema.ts
TalZaccai: Reverted to the original wording.

ts/packages/agentServer/server/src/conversationManager.ts
TalZaccai: Done.

ts/packages/vscode-shell/src/webview/globals.d.ts
TalZaccai: Yes — required. `ts/packages/vscode-shell/src/webview/globals.d.ts` provides ambient declarations for the `chat-ui/styles` and `@typeagent/completion-ui/styles.css` imports in `webview/main.ts`; without it tsc fails on those stylesheet imports.

ts/packages/agents/code/src/vscode/vscodeShellActionsSchema.ts
TalZaccai: Renamed the type to `VSCodeConversationActions` (file: `vscodeConversationActionsSchema.ts`). Left the manifest entry `code-vscode-shell` as-is for now since it's persisted in opted-in users' `user-settings.json`; happy to also rename it if you want — separate breaking-change consideration.

ts/packages/agents/code/src/vscode/editorCodeActionsSchema.ts
TalZaccai: Tried this, but the action-schema-compiler rejects `/** */` blocks with `Multi-line comments are not supported`. Reverted to a `// @deprecated …` line-comment form for now. Probably worth a follow-up to teach asc to parse JSDoc.

TalZaccai: Follow-up: matched your wording with line-comment form (`// @deprecated Use \`newFile\` in the main code schema instead.`) in a73cc770. Reason for not using `/** */`: `actionSchema/src/parser.ts:803` explicitly throws on `MultiLineCommentTrivia` for any asc-compiled action schema, because it captures leading comments as the action's prompt-time `description` metadata and only strips `// ` markers (no logic for `/**` margins). The JSDocs you'll see in `browserKnowledgeSchema.ts` / `discoveryLlmSchema.ts` are fine because those are loaded as raw text via `loadSchema` for TypeChat prompts, not parsed by asc. Worth a small follow-up to teach the parser to handle JSDoc — happy to do that separately.

ts/packages/agents/code/src/codeActionHandler.ts
TalZaccai: Refactored in 1c7f1bad — replaced the actionName-list with `action.schemaName === 'code-vscode-shell'` and extracted `executeConversationAction`. Adding a new conversation action only requires a case in that helper now; the top-level dispatcher doesn't change. Re: `NewConversationAction.actionName` typed reference: every other agent in the repo (email/list/calendar/markdown/…) also dispatches on string-literal `actionName`s, since TS types don't carry runtime values. Happy to add the `export const newConversationActionName = '…' as const` convention as a separate repo-wide pass if you'd like.

TalZaccai: Follow-up: went further in a98a82e3 — narrowed `executeConversationAction`'s parameter from `AppAction` to the `VSCodeConversationActions` union and added an `assertNever` default. This matches the typed-union pattern already used by email/calendar/markdown/weather/player/etc., and fully delivers what your original suggestion was getting at:

- Renaming an `actionName` in the schema → compile error at the case label.
- Adding a new conversation action without a case → compile error via `assertNever`.
- Each case body has fully typed access to `action.parameters` (no more `as any`).

Combined with the prior 1c7f1bad (schemaName-based top-level dispatch, matching `browserActionHandler.mts`), the dispatcher no longer carries any actionName magic strings at all.

ts/packages/agents/desktop/src/actionsSchema.ts
TalZaccai: Done in 958d56da — replaced the USE-THIS / NEVER prompt wall with the single User/Agent example you suggested.

ts/packages/agents/code/src/codeManifest.json
TalZaccai: Cleaned up in 958d56da — dropped the trigger-keyword list and the 'Does NOT handle…' negative assertions from the description. Kept the manifest entry name as `code-vscode-shell` for now since it's persisted in opted-in users' `user-settings.json`; happy to also rename it to `code-vscode-conversations` as a separate breaking-change pass if you want.

ts/packages/agents/code/src/codeActionsSchema.ts
TalZaccai: Done in 0e614b7f. Replaced `NewFileAction` with three positively-typed actions:

- `NewCodeFileAction` — `CodeFileLanguage = "html" | "css" | "json" | "python" | "javaScript" | "typeScript"`
- `NewMarkdownFileAction` — no language slot (it's implicit)
- `NewTextFileAction` — no language slot (it's implicit)

Grammar split into three rules with language-specific keywords (`codeSchema.agr`). Tests updated. The Coda extension still implements a single `newFile` handler — `codeActionHandler.ts` maps the three new action names back to the wire protocol and injects the right `language` field, so nothing on Coda's side changed.

Smoke-tested via `@config schema --priority code desktop github-cli`: `"create a markdown file with notes"` → `newMarkdownFile`, `"make a python file scratch.py"` → `newCodeFile`. Routing is clean without any anti-examples.

ts/packages/dispatcher/dispatcher/src/reasoning/claude.ts
TalZaccai: You were right — this was added because of the `newFile` cross-schema collision. With the schema split (the `NewFileAction` → `NewCodeFileAction`/`NewMarkdownFileAction`/`NewTextFileAction` split in 0e614b7f) the LLM no longer pairs `newFile` with the wrong schemaName, so this hint is no longer needed. Reverted in 027ac85a — back to the original schema-text-only error path.

ts/packages/dispatcher/dispatcher/src/reasoning/copilot.ts
TalZaccai: Reverted alongside the claude.ts revert — same root cause (cross-schema action-name collision), same fix (schema split). See 027ac85a.

ts/packages/dispatcher/dispatcher/src/translation/actionSchemaJsonTranslator.ts
TalZaccai: Removed the dotted-action-name fallback in 027ac85a. You were right that it was a hallucination guard for a problem that lives upstream — the LLM was emitting names like `code.code-editor.createFile` because the schemas had overlapping `NewFileAction`/`createFile` semantics. With the 3-way schema split (`NewCodeFileAction`/`NewMarkdownFileAction`/`NewTextFileAction`) the LLM no longer reaches for schema-qualified names; smoke testing confirmed routing works without the fallback. The injected sub-schema lookup below it (which existed before this PR) is untouched.

ts/packages/dispatcher/dispatcher/src/translation/actionSchemaJsonTranslator.ts
TalZaccai: Fixed in 027ac85a — restored the original generic `Unknown action name: ` message. The earlier version that listed `newFile`/`createFile` and `code`/`code-editor` as examples was inappropriate hardcoding of specific agent names into the dispatcher's translator.

ts/packages/chat-ui/src/chatPanel.ts
TalZaccai: Good catch — the static map will drift from the dispatcher's runtime agent set. Tracking this for the next PR: wire chat-ui's avatar lookup to consume the dispatcher's agent-icons API at runtime instead of carrying a hardcoded map. That's a focused change but it touches both the shell and chat-ui packages, so I'd rather do it in a dedicated PR than bundle it here.

ts/packages/chat-ui/src/chatPanel.ts
TalZaccai: Agreed — `chatPanel.ts` is too big and the seams (agent message rendering, input handling, layout, lifecycle) are real. Splitting it well needs its own pass since it'll touch a lot of imports and tests. Tracking for the next PR.

ts/packages/vscode-shell/src/agentServerBridge.ts
TalZaccai: Agreed — same plan as `chatPanel.ts`. `agentServerBridge.ts` has natural seams (connection lifecycle vs. dispatcher proxy vs. clientIO routing) that warrant their own files. Tracking for the next PR.

ts/packages/shell/src/renderer/src/chat/messageGroup.ts
```diff
@@ -117,12 +117,15 @@ export class MessageGroup {
...
+        ) {
             this.addStatusMessage(
                 { message: "Command completed", source: "shell" },
                 false,
             );
-        } else {
+        } else if (this.statusMessage !== undefined) {
             this.statusMessage.complete();
             this.chatView.updateScroll();
         }
```
robgruen, line 136: do we not want to scroll if there is no status message or is this code path now irrelevant?

robgruen (approved):

ts/packages/shell/src/renderer/src/chat/messageGroup.ts
TalZaccai: {
  "body": "Good catch — that was a real gap I introduced.\n\nPre-PR the chain ended with a catch-all `else { statusMessage.complete(); updateScroll(); }`. My change in 582b2e98 added the `agentMessages.length === 0` guard (to suppress the redundant \"Command completed\" status when agent bubbles are already showing) and narrowed the trailing branch to `else if (this.statusMessage !== undefined)`. That created a silent fall-through for the new common post-consolidation path (`statusMessage === undefined && agentMessages.length > 0`): no scroll, no completion. The missing scroll matters because `updateMetrics()` ran just above and may have grown the last bubble's height by inserting the metrics row, which can land below the viewport.\n\nRestored the unconditional scroll, and only call `statusMessage?.complete()` if it exists (preserves the consolidated-bubbles intent). Pushed in 001478c7.\n\nSeparately noted: `MessageContainer.complete()` also flushes pending TTS / temporary content / updates div state. Agent-message containers never get `complete()` called via this path — pre-existing behavior, but worth fixing alongside. Tracked as a follow-up todo (`next-pr-agent-message-complete`); will pick it up in the next PR rather than expand this one."
}

PR #2296, opened 2026-05-05
Improve interruption responsiveness across multiple layers

GeorgeNgMsft:
## Interrupt support: Escape/Ctrl+C now actually cancels in-flight work

Previously, pressing Escape or Ctrl+C during a command showed "Cancelled" in the UI but the underlying work — LLM network calls, streaming, agent execution — kept running until it naturally completed or timed out. This PR wires cancellation end-to-end so that abort terminates work immediately at every layer.

### Changes

- **LLM fetch termination** — The dispatcher's `AbortSignal` now reaches the actual `fetch()` call. Cancelling tears down the HTTP connection immediately, including during retry back-off waits.

- **Translation pipeline** — Cancellation is threaded through assistant selection, cache validation, TypeChat translation, action finalization, and the unknown-switcher's parallel partition translations, so the command exits at the earliest possible point instead of completing unneeded work.

- **Streaming guard** — Partial JSON chunks are dropped after cancellation, preventing agent side effects from incomplete data.

- **Agent RPC cancellation** — Out-of-process agents receive a `cancelAction` message and get a working `actionContext.abortSignal`. The shim now reuses a single `AbortController` per action so concurrent RPC entry points can't orphan the signal `executeAction` is using.

- **Utility agent** — Web fetch/search, LLM transform, and Claude task actions all honor the abort signal (Puppeteer page close + native SDK abort). Added web test cancellation coverage.

- **Early-cancel for queued commands** — A new `cancelCommandByClientId` API lets the CLI assign a UUID up front and register an `AbortController` before the command lock is acquired, so Escape works even while a command is sitting in the queue.

- **Robustness** — `AbortSignal.reason` is normalized to a real `AbortError` everywhere it's thrown, so non-Error reasons no longer slip past `isAbortError()` checks (which previously could cause endpoint-pool rotation on cancel).


hillary-mutisya (approved):

hillary-mutisya:
We should have a future PR to update chat and reasoning agents to honor the cancellation signal. There may also be other agents that trigger long-running tasks that need to be aware of cancellation

GeorgeNgMsft:
> We should have a future PR to update chat and reasoning agents to honor the cancellation signal. There may also be other agents that trigger long-running tasks that need to be aware of cancellation

Yeah, a number of agents will need to be updated, I've only addressed utility agent here as an example of an agent that spawns long running tasks that needs cancellation. However, from a UX perspective, out-of-process agents (called over RPC) will return control immediately to the user, so cancellation will still feel immediate. Agents like chat (and browser lookupAndAnswer), which is awaited by Dispatcher, will cancel on the boundary of the next turn.

PR #2304, opened 2026-05-07
[WIP] Add Discord Agent

GeorgeNgMsft:


GeorgeNgMsft:
I tried to generate this agent with the full Discord API, but it resulted in 124 actions, 660 grammars, and an indefinite reasoning loop during onboarding (didn't finish after 10+ hours). I will probably regenerate this with a smaller scenario.

PR #2332, opened 2026-05-13
Port Registrar Foundation

TalZaccai:
`agentServer` hosts many app-agents in one process, and several of those agents open their own WebSocket / HTTP listeners for out-of-process clients (Chrome extension, VS Code extension, the Visual Studio C# plugin, the onboarding helper). Today every one of those listeners hard-codes its port. That has two practical consequences:

- **Two TypeAgent installs on the same machine collide.** The second instance can't bind, and there's no good story for telling its clients where to look.
- **Changing a port is a multi-repo edit.** The agent and every external client have to move in lockstep.

We want to flip the model: agents bind `port = 0`, the OS picks a free port, the agent **registers** that port with a central registrar, and external clients **discover** the port over a well-known channel on `agentServer`.

This PR is the **foundation only** — it builds the registrar, plumbs it through the SDK, stands up the discovery channel, and ships the matching client-side discovery helper. Per-agent migrations follow in PRs 2–5.

### A shared `PortRegistrar`

The registrar is owned by the **dispatcher package** rather than `agentServer` itself, so any future host of the dispatcher (the desktop shell, a hypothetical web API) gets the same model for free. `agentServer` is just the first host that wires one up.

The registrar tracks `(agentName, role) → port` allocations, scoped by an opaque `sessionContextId` so we can clean up everything an agent registered when its session ends. Re-registering the same `(agent, role)` replaces the prior entry rather than stacking duplicates — this matches what an agent actually means when it re-binds (e.g. after a config change).

### A new SDK surface for agents

`SessionContext` gains `registerPort(role, port)` / `releasePort(id)` and exposes a stable `sessionContextId`. The same surface is mirrored in `agent-rpc` so out-of-process agents see exactly what in-process agents see.

`AppAgentManager` mints a `sessionContextId` once per `initializeSessionContext`, clears it on `closeSessionContext`, and a `finally` backstop calls `releaseAllForSession` so a partially-failed init can't leak allocations.

### A discovery channel on `agentServer`

A new WS-RPC channel named `"discovery"` is multiplexed onto the same WebSocket as the existing `agent-server` channel. It exposes one read-only call:

```
lookupPort(agentName, role?) → { port: number | null }
```

That's the only operation. Mutation (registering ports) requires being the agent itself, in-process, via `SessionContext.registerPort` — there is no remote `register`/`release` on the wire. `null` means "no such allocation"; callers don't need try/catch around every probe.

`agentServer` also calls `setAgentServerPort()` once it's listening, so **the host itself is discoverable** under the agent name `agent-server` — useful for clients that bootstrap from a different known port.

### A client-side `discoverPort()` helper

The symmetric counterpart to the discovery channel, exported via a narrow `./discovery` subpath on `@typeagent/agent-server-client`:

```ts
discoverPort(agentName, role?, { url?, timeoutMs? })
  → { kind: "found", port } | { kind: "not-registered" } | { kind: "unreachable", error }
```

The tagged result is intentional — it lets callers distinguish "agent isn't loaded yet, retry" from "agentServer isn't running, fall back to a hardcoded default for back-compat" without parsing error strings.

The reason it's a **separate subpath** rather than living on the top-level `agent-server-client` entry: extensions and service workers can't ship `fs` / `os` / `child_process` / dispatcher RPC. The `./discovery` subpath imports only `agent-rpc`, `isomorphic-ws`, and the small `agent-server-protocol` constants module. PRs 2–5 all reuse this same import.

### Idle-shutdown safety

`agentServer`'s idle-shutdown timer now bails out when `registrar.hasActiveAllocations()` is true, so cached extension clients can reconnect to a still-running server. The backstop in `closeSessionContext` ensures the registrar drains naturally on session close, so this guard never wedges shutdown indefinitely.

### Origin-allowlist hook on `webSocketChannelServer`

`webSocketChannelServer` accepts an optional `originAllowlist` (case-insensitive, supports a trailing `*` for prefix-match like `chrome-extension://*`; native CLI clients with no Origin header are always allowed). The hook is in place so future hardening doesn't require another protocol change.

### Constants instead of `8999` everywhere

`AGENT_SERVER_DEFAULT_PORT = 8999` and `AGENT_SERVER_DEFAULT_URL` now live in `agent-server-protocol`. All the places that previously hard-coded `8999` (CLI commands, the shell, vscode-shell, the Visual Studio webview, the browser extension service worker, `uriHandler`, `commandExecutor`, `agentServer/{stop,status}`) now import the constant. Behaviorally identical; just removes a class of "I changed it in one place but not the other" bugs.

## A wire-format change worth calling out

`ContextParams` on `agent-rpc` gains a **required** `sessionContextId` field. This is safe because rpc client and server always build and ship together within this monorepo, but it's worth flagging for anyone watching for compatibility-affecting changes.

## Validation

- Full `pnpm run build` from `ts/` — green.
- `pnpm run jest-esm --testPathPattern="portRegistrar|sessionContext"` — **27/27 passing** (registrar + session lifecycle).
- `pnpm --filter @typeagent/agent-server-client test` — **4/4 passing** for `discoverPort`. These spin up a real `ws` server speaking the `agent-rpc` channel framing and exercise `found`, `not-registered`, `unreachable` (server bound and immediately closed), and timeout (server accepts but never resolves) — the missing wire-protocol coverage for the discovery channel.
- Manual smoke: `agentServer` starts on the default port, idle-shutdown defers while an allocation is held, and releases the moment `closeSessionContext` runs.

The PR is four commits, each independently buildable and reviewable:

1. **`feat(dispatcher): add PortRegistrar class with unit tests`** — pure data structure + tests. Read this first to internalize the model; nothing else depends on the rest of the codebase.
2. **`refactor(dispatcher): integrate PortRegistrar into AppAgentManager + SDK`** — wires the registrar into the dispatcher's session lifecycle and exposes the agent-facing API. The interesting bits are the `sessionContextId` lifecycle in `appAgentManager.ts` and the matching surface on `agent-rpc`.
3. **`agentServer: add discovery channel + consolidate AGENT_SERVER_PORT`** — the host wiring. The discovery channel handler in `server.ts` is small (one RPC method); most of the diff size is the mechanical `8999 → constant` replacement across CLI/shell/extensions.
4. **`agent-server-client: add discoverPort helper + wire-protocol tests`** — symmetric client of commit 3. Small standalone helper + 4 integration tests against a real `ws` server speaking the actual framing. Bootstraps Jest in the package using the standard repo convention.

## Follow-up PRs

| PR | Scope |
|----|-------|
| 2 | Migrate `code` agent (and trivially `localView`) — first consumer of `discoverPort` |
| 3 | Migrate `browser` agent service worker |


ts/packages/agentServer/client/src/discovery.ts
```diff
@@ -0,0 +1,174 @@
...
+     * override (e.g. `AGENT_SERVER_URL`) should resolve it themselves
+     * and pass the result here.
+     */
+    url?: string;
+    /** Per-attempt timeout in milliseconds. Default 5_000. */
+    timeoutMs?: number;
+};
+
+/** What `discoverPort` returns. */
+export type DiscoverPortResult =
```
robgruen, line 50: Do we want the notion of a transient port (i.e. one that can come and go i.e. not guaranteed to be up)?

robgruen: This isn't the place for that, just asking here since it just came to me.

ts/packages/agentServer/client/src/discovery.ts
```diff
@@ -0,0 +1,174 @@
...
+ * fall back to a hardcoded default for back-compat" without parsing
+ * error messages.
+ *
+ * Closes the underlying WS after the lookup completes (success or
+ * failure) — discovery connections are intentionally short-lived to
+ * keep the agentServer's idle-shutdown timer honest.
+ */
+export async function discoverPort(
+    agentName: string,
+    role?: string,
```
robgruen, line 86: Should this be an enum?

ts/packages/agentServer/server/src/server.ts
```diff
@@ -155,6 +160,14 @@ async function main() {
         configIdx !== -1 ? process.argv[configIdx + 1] : undefined;
 
     debugStartup("creating conversation manager (will lockInstanceDir)");
+    // Single PortRegistrar shared across every conversation in this
+    // process. Lets external clients (browser extension, VS Code, CLI)
+    // discover any agent's port via the discovery channel regardless of
+    // which conversation that agent is loaded into. Standalone hosts
+    // (shell, CLI dispatcher) skip this and let each dispatcher mint
+    // its own — see DispatcherOptions.portRegistrar in agent-dispatcher.
+    const portRegistrar = new PortRegistrar();
```
robgruen, line 169: Should this then be a singleton if we ever just want one of them?

ts/packages/cli/src/slashCommands.ts
```diff
@@ -184,7 +187,7 @@ const slashCommands: SlashCommand[] = [
         name: "shutdown",
         description: "Shut down the agent server",
         handler: async () => {
-            const port = serverPort ?? 8999;
+            const port = serverPort ?? AGENT_SERVER_DEFAULT_PORT;
```
robgruen, line 190: dang, this was ALL OVER THE PLACE!

ts/packages/dispatcher/dispatcher/src/context/portRegistrar.ts
```diff
@@ -0,0 +1,265 @@
...
+ * channel) can look it up.
+ *
+ * Lifetime: one instance per dispatcher (i.e. per agentServer process,
+ * per shell, etc.). Not persisted across restarts — agents re-bind and
+ * re-register on each run.
+ *
+ * Thread-safety: Node single-threaded; the registrar is mutated only on
+ * the event-loop thread. No locking required.
+ */
+export class PortRegistrar {
```
robgruen: Should this thing implement an interface such that others can provide their own port registration mechanism?  Couldn't hurt to define the interface and export it on the off chance.

ts/packages/dispatcher/dispatcher/src/context/portRegistrar.ts
```diff
@@ -0,0 +1,265 @@
...
+     * and refusing to record the port would just hide the listener from
+     * lookups.
+     */
+    public register(
+        agentName: string,
+        role: string,
+        port: number,
+        sessionContextId: string,
+    ): RegistrationId {
+        if (!Number.isInteger(port) || port < 0 || port > 65535) {
```
robgruen, line 124: Should also list the agent-server's own port here.

ts/packages/dispatcher/dispatcher/src/context/portRegistrar.ts
```diff
@@ -0,0 +1,265 @@
...
+        if (
+            this.agentServerPort !== undefined &&
+            port === this.agentServerPort
+        ) {
+            debugWarn(
+                `${agentName}/${role} registered the agentServer's own port ${port}; this is almost certainly a hard-coded mistake — pass 0 to bind`,
+            );
+        }
+
+        const tripleKey = this.makeTripleKey(agentName, role, sessionContextId);
```
robgruen: instead of a triple key, couldn't this be an object that itself is the key?  Would be easier to extend if we add a fourth property.

ts/packages/dispatcher/dispatcher/src/context/portRegistrar.ts
```diff
@@ -0,0 +1,265 @@
...
+            `register ${agentName}/${role} session=${sessionContextId} port=${port} regId=${regId}`,
+        );
+        return regId;
+    }
+
+    /**
+     * Remove a single allocation by its registration id. Idempotent: a
+     * release of an unknown id is a no-op.
+     */
+    public release(regId: RegistrationId): void {
```
robgruen: Probably need a way to ensure that agents only release their own ports.

robgruen (approved):

ts/packages/cli/src/slashCommands.ts
TalZaccai: YUUUUP

ts/packages/agentServer/protocol/src/protocol.ts
```diff
@@ -67,6 +67,44 @@ export const DefaultUserIdentity: UserIdentity = {
...
+export type DiscoveryInvokeFunctions = {
+    /**
+     * Look up the port currently registered for `(agentName, role)`.
+     * Returns `null` (not undefined) so the JSON-RPC response is always a
+     * defined value; callers should treat null as "no allocation found,
+     * try again later" rather than a hard error.
+     */
+    lookupPort: (param: {
+        agentName: string;
+        role: string;
```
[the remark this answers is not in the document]
TalZaccai: Already in place — 
ole?: string at protocol.ts:120 (optional). Looks like the bot didn't pick up the existing optional marker. Closing as stale.

ts/packages/agentRpc/src/client.ts
```diff
@@ -184,6 +185,9 @@ export async function createAgentRpcClient(
 ) {
     const channel = channelProvider.createChannel(`agent:${name}`);
     const contextMap = createObjectMap<SessionContext<ShimContext>>();
+    // Tracks port registration handles returned by sessionContext.registerPort
+    // so the out-of-process agent can release them via the regId we sent back.
+    const registrationHandles = new Map<string, { release: () => void }>();
```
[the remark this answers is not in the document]
TalZaccai: Already addressed in this PR — see `client.ts:783-806` where `closeAgentContext` walks `regIdsByContext[contextId]` and releases any handles the out-of-process agent forgot. `regIdsByContext` is populated in the `registerPort` handler (line 349-354) and pruned in `releasePort`. Mirrors the dispatcher-side `releaseAllForSession` backstop.

ts/packages/agentServer/client/src/discovery.ts
TalZaccai: Good question — and the docs were thin here, which made an enum a reasonable thing to ask about. Updated the JSDoc on `discoverPort`, `registerPort`, and `lookupPort` (in the new commit) to make the namespace ownership explicit:

- Roles are **agent-defined** free-form strings. The registrar / discovery / SDK layer is intentionally generic transport and doesn't enumerate valid values.
- Each agent owns its own role namespace and **should publish role constants** for in-process and out-of-process callers (e.g. `code` could export `CODE_ROLES = { default: 'default', debug: 'debug' } as const`). That gives callers the type-safety benefit of an enum without coupling.
- A central enum would force every new role to amend a shared file across packages and create a discovery -> all-agents dependency cycle (this layer can't import every agent that exposes a port).

ts/packages/agentServer/server/src/server.ts
TalZaccai: It already is, in practice — there's exactly one `PortRegistrar` per host process: the agent-server constructs one in `server.ts:169` and threads it through `DispatcherOptions.portRegistrar` so every conversation in that process shares the same instance. Standalone hosts (shell, CLI dispatcher) each mint their own (`commandHandlerContext.ts:589`).

I deliberately didn't make it a module-level singleton because:
1. Tests construct their own per-test-case (see `portRegistrar.spec.ts`) — a singleton would force test isolation hacks.
2. A future host that runs multiple isolated dispatchers in one process (multi-tenant testbed) would step on itself with a true singleton.

The `IPortRegistrar` interface added in this commit makes the substitutable-instance pattern explicit.

ts/packages/dispatcher/dispatcher/src/context/portRegistrar.ts
TalZaccai: Done in b06f1d43 — `IPortRegistrar` interface extracted in `portRegistrar.ts` and consumers (`AppAgentManager`, `CommandHandlerContext`, `DispatcherOptions`) now depend on the interface. Concrete `PortRegistrar` class still exported for the agent-server's construction site.

ts/packages/dispatcher/dispatcher/src/context/portRegistrar.ts
TalZaccai: Done in b06f1d43 — the agent-server's port is now registered as a regular allocation under `AGENT_SERVER_DISCOVERY_NAME` with a synthetic `SYSTEM_SESSION_CONTEXT_ID`, so it shows up in `list()` / `lookup()` / `hasActiveAllocations()` like any other. The `setAgentServerPort`/`getAgentServerPort` side fields and the special-case in the discovery handler are gone. `releaseAllForSession` refuses to act on the system id so a stray session close can't cull it.

ts/packages/dispatcher/dispatcher/src/context/portRegistrar.ts
TalZaccai: Done in b06f1d43 — replaced the inline string-concat with a `keyOf(allocation)` helper that takes an `Allocation`-shaped object. Field order is now enforced by the type rather than by call-site discipline, and there's a single place to look at the key encoding (still `\u0000`-delimited since UUIDs and TS-identifier-shaped names can't contain it).

ts/packages/dispatcher/dispatcher/src/context/portRegistrar.ts
TalZaccai: Done in b06f1d43 — `release(regId, sessionContextId?)` now verifies ownership when the second arg is supplied. The SDK shim (`sessionContext.ts`) and the agentRpc release path (`client.ts:357-372`) both pass the owner so a buggy or compromised agent can't cancel another agent's registration by guessing/replaying a regId. Mismatches log under the warn debug namespace and are silently dropped.

ts/packages/agentServer/server/src/server.ts
```diff
@@ -416,9 +440,32 @@ async function main() {
...
+                "agent-server:discovery",
+                channelProvider.createChannel(DiscoveryChannelName),
+                discoveryFunctions,
+            );
         },
     );
 
+    // Tell the registrar which port we're bound to so it can warn agents
+    // that try to register the same one (a foot-gun: agent-server's WS
+    // would silently shadow the agent's listener).
```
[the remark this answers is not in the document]
TalZaccai: Addressed in b06f1d43 — agent-server's own port is now registered as a regular allocation in the registrar (`server.ts`: `portRegistrar.register(AGENT_SERVER_DISCOVERY_NAME, 'default', port, SYSTEM_SESSION_CONTEXT_ID)`). The discovery `lookupPort` handler no longer needs the special-case — it just calls `portRegistrar.lookup()` for everyone.

ts/packages/agentSdk/src/agentInterface.ts
```diff
@@ -226,12 +235,42 @@ export interface SessionContext<T = unknown> {
...
+     * this agent — e.g. `"ws-bridge"`, `"http-debug"`. Use distinct
+     * roles when an agent exposes multiple listeners.
+     *
+     * Returns a `release()` callback the agent should invoke when the
+     * listener is torn down. Forgetting to release is non-fatal — the
+     * dispatcher releases all of an agent's allocations as a backstop
+     * when the agent's session context closes — but explicit release
+     * keeps lookups accurate while the agent is still alive.
+     */
+    registerPort(role: string, port: number): { release: () => void };
```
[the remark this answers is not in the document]
TalZaccai: The code is correct — `registerPort(role: string, port: number)` (`agentInterface.ts:331`). The PR description had it as `registerPort(port, role?)` which is wrong; updating the description now.

PR #2334, opened 2026-05-13
Port Registrar: migrate code agent + Coda extension to dynamic ports

TalZaccai:
> **Depends on PR #2332.** Branched off `talzacc/port-registrar-foundation` and must merge after it. The new SDK surface (`SessionContext.registerPort`, the `discovery` WS-RPC channel, the `discoverPort()` client helper, the shared port constants) all come from previous PR.

## Why

PR #2332 stood up the registrar, the discovery channel, the client-side `discoverPort()` helper, and the SDK. But no shipping agent uses any of it yet — every agent still hard-codes its port. This PR is the first concrete migration and proves the model end-to-end: an in-process agent binding `port = 0`, registering via the SDK, and an external client discovering the live port over the discovery channel before connecting.

### `code` agent: dynamic port + per-session registration

`CodeAgentWebSocketServer` is now an async `start(port = 0)` factory rather than a constructor. It resolves only after the `listening` event fires, which lets us read the OS-assigned port reliably, and `close()` returns a Promise so callers can serialize a `disable → re-enable` cycle (matters when an operator pins a fixed port via `CODE_WEBSOCKET_PORT`).

`updateCodeContext` registers the server's actual bound port via `sessionContext.registerPort("default", server.port)` on every enable, and stores the registration handle on the per-session `agentContext` so `disable` can release it. The shared module-scoped server (refcounted across sessions) is preserved — only the registrations are per-session. This is intentional: the registrar already supports multiple registrations for the same `(agent, role)` and `lookup` returns the most recent, so we don't need an ownership-transfer state machine. If the dispatcher tears a session down hard, `releaseAllForSession` is the crash-safety backstop.

Race handling worth noting:

- A `sharedStartingPromise` lock keeps two concurrent enables from racing each other into a double-bind.
- A `sharedClosingPromise` lock makes a rapid disable→enable wait for the prior close to land before re-binding (only matters under a fixed-port override; with `port = 0` the OS would just hand us a different port).
- On bind failure during enable the schema is removed from `agentContext.enabled` so the agent's view of state stays consistent, and refcount is left untouched (it only increments after a successful register).

### `readiness.ts`: port becomes optional

The readiness probe and setup card now treat port as `number | undefined`. The setup card and NOT-RUNNING messaging omit the `:port` suffix entirely when no override is set, since "the port is whatever the OS gave us, ask the registrar" is the new normal. `resolveCodePort` was renamed to `resolveCodePortOverride` to make its env-var semantics explicit (`CODE_WEBSOCKET_PORT` is now strictly an escape hatch).

### Coda VS Code extension: discovery-then-connect

`coda/src/webSocket.ts` now does:

1. If `CODE_WEBSOCKET_HOST` is set, use it as a hard override (unchanged escape hatch).
2. Otherwise call `discoverPort("code")` (from `@typeagent/agent-server-client/discovery`) against the agentServer's known port.
3. On `unreachable` (no agentServer), fall back to `ws://localhost:8082` for back-compat with old in-process agentServers that haven't been updated yet.
4. On `not-registered`, surface as a connect failure (the agent simply isn't loaded right now).

The reconnect loop (`wsConnect.ts`) now guards against the multi-interval leak in the prior implementation — every disconnect previously spawned a fresh `setInterval`, so a flapping connection would stack timers — and re-runs discovery on every retry, since the code agent's port can change across agentServer restarts now that it's OS-assigned.

`createWebSocket` had to become async because the discovery handshake is async; one call site in `wsConnect.ts` was updated accordingly.

## Validation

- Full `pnpm run build` from `ts/` — green.
- `pnpm run lint` from `ts/` — clean.
- `pnpm --filter code-agent test` — **18/18 passing** (13 readiness tests updated for the new optional-port type, 5 new integration tests for the shared-server lifecycle exercising real WebSocket binds on `port = 0`).

This PR is one squashed commit, but reads naturally in two layers:

1. **`agents/code/src/{codeAgentWebSocketServer,codeActionHandler,readiness}.ts` + tests** — agent-side migration. The interesting bits are the async start/close, the `ensureSharedServer` race locks, and the per-session registration handle on `agentContext`.
2. **`coda/src/{webSocket,wsConnect}.ts`** — extension-side consumer. Demonstrates the discovery-then-connect pattern (using `discoverPort()` from PR 1) and the reconnect-loop guard.

## Follow-up PRs

| PR | Scope |
|----|-------|
| 3 | Migrate `browser` agent service worker (uses the same `discoverPort`) |


ts/packages/coda/src/webSocket.ts
```diff
@@ -11,17 +12,57 @@ export type WebSocketMessage = {
...
+            "code agent not yet registered with agent-server; will retry.",
+        );
+        return undefined;
+    }
+    // Discovery unreachable (agentServer not running). Reconnect loop
+    // will retry.
+    console.log(
+        `Discovery channel unreachable (${result.error.message}); will retry.`,
+    );
+    return undefined;
```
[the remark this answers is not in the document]
TalZaccai: Intentional. This is a research project with no pre-discovery clients in the wild to support — the back-compat fallback was removed in commit e7474262 at the maintainer's request. When discovery is unreachable we now log `Discovery channel unreachable (...); will retry.` and let the existing 5s reconnect loop in wsConnect.ts retry, which recovers automatically once the agentServer is back up. Verified end-to-end in the manual smoke test.

ts/packages/coda/src/webSocket.ts
```diff
@@ -11,17 +12,57 @@ export type WebSocketMessage = {
...
     channel: string,
     role: string,
     clientId?: string,
 ) {
-    return new Promise<WebSocket | undefined>((resolve, reject) => {
-        let endpoint =
-            process.env["CODE_WEBSOCKET_HOST"] ?? "ws://localhost:8082";
-        endpoint += `?channel=${channel}&role=${role}`;
+    return new Promise<WebSocket | undefined>(async (resolve) => {
+        const base = await resolveCodeEndpoint();
```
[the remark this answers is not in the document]
TalZaccai: Good catch — fixed in commit a0c62b8a. Refactored `createWebSocket` so the awaits happen at the function's top level (no async Promise executor) and wrapped `new WebSocket()` in try/catch to handle the invalid-`CODE_WEBSOCKET_HOST` case. `initializeWS` now reliably falls into the reconnect loop on either failure mode.

ts/packages/agents/code/test/codeUpdateContext.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    afterEach(async () => {
+        // Defensive: if a test left the shared server up, calling
+        // updateAgentContext(false, ...) on a synthetic session won't
+        // close it (refcount stays positive). Reach into the module's
+        // close path by exhausting any sessions via the public API —
+        // but since we control all stubs in each test, the per-test
+        // cleanup handles it. Leaving this hook as a placeholder.
+    });
```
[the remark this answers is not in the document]
TalZaccai: Fixed in commit a0c62b8a. `newSession()` now registers each `StubContext` in a module-scoped `liveSessions` set, and `afterEach` force-disables every entry (best-effort; swallows per-session errors so others still tear down). It also asserts `getSharedCodePort() === undefined` after cleanup, so a real leak fails the test loudly instead of silently contaminating later ones.

ts/packages/agents/code/src/codeActionHandler.ts
robgruen, line 1: The "reserve a port" code seems like it will be a pattern repeated by agents...if thats the case, are we planning to put this in the AgentSDK?

ts/packages/agents/code/src/readiness.ts
```diff
@@ -102,16 +107,21 @@ export async function whichExists(tool: string): Promise<boolean> {
...
+export function resolveCodePortOverride(
+    env: NodeJS.ProcessEnv,
+): number | undefined {
     const raw = env.CODE_WEBSOCKET_PORT;
-    if (raw !== undefined) {
-        const n = parseInt(raw);
-        if (Number.isFinite(n) && n > 0) return n;
-    }
-    return 8082;
+    if (raw === undefined) return undefined;
```
robgruen, line 124: Maybe log that we are doing a port override.

ts/packages/coda/src/webSocket.ts
```diff
@@ -11,31 +12,80 @@ export type WebSocketMessage = {
...
         if (clientId) {
-            endpoint += `clientId=${clientId}`;
+            endpoint += `&clientId=${clientId}`;
         }
-
-        const webSocket = new WebSocket(endpoint);
-
-        webSocket.onopen = (event: object) => {
+    } catch (error) {
+        console.error("Error resolving code agent endpoint:", error);
```
robgruen, line 66: debug() log, or both?

ts/packages/coda/src/webSocket.ts
```diff
@@ -11,31 +12,80 @@ export type WebSocketMessage = {
...
+        try {
+            webSocket = new WebSocket(endpoint);
+        } catch (error) {
+            // e.g. invalid URL from CODE_WEBSOCKET_HOST override.
+            console.error("Error constructing WebSocket:", error);
+            resolve(undefined);
+            return;
+        }
+        webSocket.onopen = () => {
             console.log("websocket open");
```
robgruen, line 80: same same, for the below methods as well.  debug() or both debug + console.log

robgruen (approved):

ts/packages/agents/code/src/codeActionHandler.ts
TalZaccai: Good question. The actual port-reservation API `SessionContext.registerPort(role, port)` is already in the AgentSDK ([agentInterface.ts:336](https://github.com/microsoft/TypeAgent/blob/talzacc/code-and-localview-migration/ts/packages/agentSdk/src/agentInterface.ts#L336)) and is what every migrated agent calls. What's per-agent here is the 3-line ceremony around it: `bind(0)` -> read the OS-assigned port off the listener -> register. We deliberately kept that per-agent because the listener types differ across migrations - `code` and `browser` use a `ws` server, `browser` view-server forks a child process, the upcoming `visualStudio` migration will use a different webview transport. A shared SDK helper would have to be either transport-specific (defeating the point) or so abstract it'd just be `register(port)` -> i.e. the existing `registerPort()`. If we end up with three or more agents using the exact same WS pattern, extracting an `sdk/helpers/wsListener` helper would be a clean follow-up. Filed as a future consideration in the design doc, no work for this PR.

ts/packages/agents/code/src/readiness.ts
TalZaccai: Done in commit a20e8135. `resolveCodePortOverride` now logs `CODE_WEBSOCKET_PORT override active: using port N (set in environment)` when the override is honored, and `CODE_WEBSOCKET_PORT override ignored: invalid value <raw>; falling back to OS-assigned port` when a typo'd value falls through. Both go through the existing `typeagent:code` debug namespace.

ts/packages/coda/src/webSocket.ts
TalZaccai: Intentional `console.*` here - it matches the existing convention across the entire `packages/coda` extension (every file in coda uses `console.log/error/warn` exclusively; the package never adopted the `debug` package). Keeping the new code consistent with the surrounding 12 files rather than introducing a new pattern just for `webSocket.ts`. If we want to migrate the coda extension to `debug()` it should be a single sweep across all `console.*` call sites, not a partial conversion in this PR.

ts/packages/coda/src/webSocket.ts
TalZaccai: Same answer as the comment above on line 66 - keeping `console.*` to match the rest of the coda package. Happy to do a full `console -> debug` sweep of `packages/coda` as a separate PR if that's the direction we want.

PR #2374, opened 2026-05-20
Port Registrar: visualStudio agent migration + C# discovery client

TalZaccai:
## Why

The `visualStudio` bridge bound a hardcoded port (`5680`) and the VSIX dialed the matching hardcoded URI. This blocks the design goal of OS-assigned ephemeral ports brokered through `PortRegistrar`.

## What changes

**TS bridge (`visualStudioActionHandler.ts`)** — rewritten to mirror the `code` and `browser` handlers:

- Async `VisualStudioBridge.start(port = 0)` factory; resolves on `listening`, rejects on first `error`.
- `verifyClient` Origin gate via a new `originAllowlist.ts` (loopback + absent/`null`; the C# `ClientWebSocket` sends no Origin header).
- Module-scoped shared bridge with refcount + start/close serialization, so rapid enable/disable cycles under a fixed-port override don't race into `EADDRINUSE`.
- Per-session `registerPort("default", bridge.port)`, released in both `updateAgentContext(false)` and `closeAgentContext`.
- `VISUALSTUDIO_BRIDGE_PORT` escape hatch, mirroring the other agents.
- **Multi-client + `@system ports` integration** — bridge tracks all connected VS instances via `Map<clientId, WebSocket>` (instead of silently displacing on second connect); per-request `clientId` so a disconnect only fails that client's pending RPCs. Emits `onClientCountChanged` to the session context, surfaced via `notifyClientCountChanged` with a primary-session fanout so shared sessions don't double-count. `send()` routes to the most-recently-connected OPEN client (preserves legacy last-wins behavior).

**C# bridge** — VSIX is .NET Framework 4.7.2, can't reuse the TS discovery helper:

- New `Bridge/BridgeDiscovery.cs` (~190 LOC, single public method) speaks the agent-rpc discovery protocol over `ClientWebSocket`. Never throws — returns the discovered port or a fallback.
- `AgentBridgeClient.cs` drops the hardcoded URI; resolves the port on every reconnect, so a running VS adapts to an agent-server restart without an extension reload.
- Three env knobs (documented in README): `AGENT_SERVER_PORT` (default `8999`), `TYPEAGENT_VS_USE_DISCOVERY` (default on), `TYPEAGENT_VS_FALLBACK_PORT` (default `5680`).

**Docs** — README diagram + "Port discovery" / troubleshooting sections updated.

## Validation

- `pnpm --filter visualstudio-agent build` — green.
- Prettier — clean.
- C# — compiles against the existing csproj; no new NuGet deps.

## Notes

- Role is `"default"` (not `"ws-bridge"` from the design doc) to match `code` / `browser` and keep `@system ports` consistent.
- `originAllowlist.ts` is deliberately duplicated rather than shared — different scheme prefixes per agent, ~30 LOC.
- No new tests; this package has no existing TS or C# test harness. Happy to add layer-6/layer-8 coverage as a follow-up.

## Reading order

1. `src/originAllowlist.ts` — origin policy.
2. `src/visualStudioActionHandler.ts` — TS bridge rewrite.
3. `host/csharp/Bridge/BridgeDiscovery.cs` — new discovery client.
4. `host/csharp/Bridge/AgentBridgeClient.cs` — hardcoded URI removed.
5. `README.md` — new env vars, discovery flow.


ts/packages/agents/visualStudio/src/visualStudioActionHandler.ts
```diff
@@ -31,49 +41,138 @@ type BridgeResponse = {
...
             });
             ws.on("close", () => {
-                console.log("[visualStudio] host plugin disconnected");
+                debug("host plugin disconnected");
                 this.client = undefined;
             });
             ws.on("error", (err) => {
-                console.error("[visualStudio] host plugin socket error:", err);
+                debug("host plugin socket error:", err);
             });
```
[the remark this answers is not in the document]
TalZaccai: Fixed in ba4ae2055. The `pending` map now stores `{ resolve, timer }` and is drained via a shared `failPending(error)` helper on `close`, `error`, and `stop()`. Each `send()` also installs a per-request timer (configurable via `VISUALSTUDIO_BRIDGE_SEND_TIMEOUT_MS`, default 30s), so awaits no longer hang forever and the map cannot grow unbounded across reconnects.

dotnet/visualStudioTypeAgent/Bridge/BridgeDiscovery.cs
```diff
@@ -0,0 +1,170 @@
...
+            }
+            return discovered;
+        }
+
+        private static int GetAgentServerPort()
+        {
+            string? raw = Environment.GetEnvironmentVariable(AgentServerPortEnv);
+            if (int.TryParse(raw, out int p) && p > 0 && p <= 65535)
+            {
+                return p;
```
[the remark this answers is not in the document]
TalZaccai: Already addressed in adf142684 (pushed prior to this review). `BridgeDiscovery.ResolvePortAsync` now returns `int?` and propagates `null` to `AgentBridgeClient`, which surfaces a clear "agent not registered" error instead of silently dialing 5680.

ts/packages/agents/visualStudio/src/originAllowlist.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Allowed:
+ *  - **No Origin header** — the C# `ClientWebSocket` doesn't set one.
+ *  - `http(s)://localhost(:port)`, `http(s)://127.0.0.1(:port)`,
+ *    `http(s)://[::1](:port)` — loopback web clients (manual debugging,
+ *    future webview consumers). The server already binds to loopback, so
+ *    this is OS-level restricted as well.
+ *
+ * Anything else (including arbitrary `https://example.com`) is rejected
+ * with HTTP 403 before the `connection` event fires.
+ *
```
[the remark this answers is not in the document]
TalZaccai: Fixed in ba4ae2055: the server now actually binds to loopback (`host: '127.0.0.1'`), so the doc comment is now accurate.

ts/packages/agents/visualStudio/src/visualStudioActionHandler.ts
```diff
@@ -31,49 +41,138 @@ type BridgeResponse = {
...
+                    const origin = info.req.headers.origin as
+                        | string
+                        | undefined;
+                    if (isAllowedAgentOrigin(origin)) {
+                        cb(true);
+                    } else {
+                        debug(`Rejecting WS upgrade from origin ${origin}`);
+                        cb(false, 403, "Origin not allowed");
+                    }
+                },
```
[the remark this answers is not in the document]
TalZaccai: Fixed in ba4ae2055: `WebSocketServer` now binds to `host: '127.0.0.1'`, so even though the C# `ClientWebSocket` doesn't send an Origin header (and the allowlist accepts no-Origin), the listener is unreachable from the LAN.

dotnet/visualStudioTypeAgent/Bridge/BridgeDiscovery.cs
```diff
@@ -0,0 +1,170 @@
...
+    ///       "message": { "type": "invokeResult", "callId": N,
+    ///                    "result": { "port": &lt;int|null&gt; } } }
+    ///
+    /// Returns the discovered port, or null when the agent isn't yet
+    /// registered with the agent-server (transient — caller should retry).
+    /// Throws on transport failure so the outer reconnect loop can apply
+    /// its own retry/backoff. There is intentionally no hardcoded fallback
+    /// port — the migrated TS clients (browser, code, coda) all return
+    /// "undefined" on discovery failure and rely on the reconnect loop;
+    /// dialing a stale well-known port would just connect to nothing.
```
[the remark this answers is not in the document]
TalZaccai: Already addressed in adf142684. `ResolvePortAsync` now throws on transport failure (matches the XML doc); only the "not registered" case returns `null`.

ts/pnpm-lock.yaml
```diff
@@ -3168,13 +3168,19 @@ importers:
...
     devDependencies:
       '@typeagent/action-schema-compiler':
         specifier: workspace:*
         version: link:../../actionSchemaCompiler
+      '@types/debug':
+        specifier: ^4.1.12
+        version: 4.1.12
       '@types/ws':
         specifier: ^8.5.10
         version: 8.18.1
```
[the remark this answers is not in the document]
TalZaccai: Regenerated in ba4ae2055. The diff vs main is now +27/-202 lines — the additions are `debug` + `@types/debug` for the visualStudio importer; the deletions are pnpm consolidating two pre-existing `@types/node` resolutions (22.15.18 + 22.19.19) into one when it re-resolved the workspace. No unrelated downgrades remain.

ts/packages/agents/visualStudio/host/csharp/Bridge/BridgeDiscovery.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// port, or null when the agent has not yet registered (transient
+        /// — caller should retry on its reconnect loop).
+        /// Throws on transport failure (agent-server unreachable, timeout,
+        /// malformed response) so the caller can log and retry.
+        /// </summary>
+        public static async Task<int?> ResolveBridgePortAsync(CancellationToken cancellation)
+        {
+            int agentServerPort = GetAgentServerPort();
+            int? discovered = await LookupPortAsync(agentServerPort, cancellation).ConfigureAwait(false);
+            if (discovered is int p)
```
robgruen: nit: uint

ts/packages/agents/visualStudio/host/csharp/Bridge/BridgeDiscovery.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (int.TryParse(raw, out int p) && p > 0 && p <= 65535)
+            {
+                return p;
+            }
+            return DefaultAgentServerPort;
+        }
+
+        private static async Task<int?> LookupPortAsync(int agentServerPort, CancellationToken cancellation)
+        {
+            var uri = new Uri($"ws://localhost:{agentServerPort}/");
```
robgruen: ws://localhost string repeated frequently.  make it a const for now, in the future URL may be configurable.

dotnet/visualStudioTypeAgent/Bridge/BridgeDiscovery.cs
```diff
@@ -0,0 +1,170 @@
...
+            try
+            {
+                await ws.CloseAsync(WebSocketCloseStatus.NormalClosure, null, CancellationToken.None).ConfigureAwait(false);
+            }
+            catch
+            {
+                // Best-effort close — the response is already in hand.
+            }
+
+            var root = JObject.Parse(responseText);
```
robgruen: don't we have some styleCOP we can reuse?  I think we force {} everywhere else.

dotnet/visualStudioTypeAgent/Bridge/BridgeDiscovery.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (inner.Value<int?>("callId") != callId) return null;
+            var result = inner["result"] as JObject;
+            if (result == null) return null;
+            // `port` is `int|null`; JObject returns null cleanly for both.
+            return result.Value<int?>("port");
+        }
+
+        private static async Task<string> ReceiveFullMessageAsync(ClientWebSocket ws, CancellationToken cancellation)
+        {
+            var buffer = new ArraySegment<byte>(new byte[16 * 1024]);
```
robgruen: Arbitrary limit?  We should probably handle this better (maybe by including the message size or something).

ts/packages/agents/visualStudio/src/visualStudioActionHandler.ts
```diff
@@ -31,84 +41,428 @@ type BridgeResponse = {
 };
 
 class VisualStudioBridge {
```
robgruen, line 43: Is this the same pattern from the other agents?  Should this be a standard implementation across the agents that then takes in an interface in construction that it can call when something interesting happens?  Seems like that should go in the same library as the originAllowList that keeps cropping up.

robgruen (approved):

ts/packages/agents/visualStudio/host/csharp/Bridge/BridgeDiscovery.cs
TalZaccai: Fixed in b8b9e8ded92ec734c282da1587fe74a4d0b63bd7. The discovery path now uses `uint` for port values (parsing, return type, override env). The Socket/Uri boundary in `AgentBridgeClient.ConnectAndReceiveAsync` still takes `int` so there is one explicit cast at that boundary.

ts/packages/agents/visualStudio/host/csharp/Bridge/BridgeDiscovery.cs
TalZaccai: Fixed in b8b9e8ded. Extracted to `AgentServerHost` const at the top of the class. When the host becomes configurable later, the build target is the single const + `GetAgentServerPort` (mirroring how the port already works).

dotnet/visualStudioTypeAgent/Bridge/BridgeDiscovery.cs
TalZaccai: Fixed in b8b9e8ded. Wrapped every single-statement `if` body in this file in braces. We don't have StyleCop wired into the C# project yet (would need to be added to `VisualStudioTypeAgent.csproj` with a ruleset), so for now this is best-effort consistency. Happy to file a follow-up for actually wiring StyleCop if you want enforcement.

dotnet/visualStudioTypeAgent/Bridge/BridgeDiscovery.cs
TalZaccai: Fixed in b8b9e8ded. The 16KB `buffer` is the per-`ReceiveAsync` chunk size, not a hard message cap (the loop drains until `EndOfMessage`). Added an explicit `MaxDiscoveryResponseBytes = 64 * 1024` total-payload cap that throws `InvalidOperationException` if the peer streams beyond it, plus a clarifying comment so the buffer size isn't mistaken for the limit. The discovery envelope is ~100 bytes in practice so 64KB is generous; happy to tune if needed.

ts/packages/agents/visualStudio/src/visualStudioActionHandler.ts
TalZaccai: Agreed this duplication is real ΓÇö `originAllowlist` is already copy-pasted in 4+ agents and the bridge `pending`-map / `send-timeout` / shared-singleton pattern is now duplicated across multiple agent scenarios. The right answer is a `@typeagent/agent-bridge-utils` (or similar) library that exposes:

1. `createAgentOriginAllowlist` (already extracted, just not packaged) plus a standard upgrade-handler that calls `isAllowedAgentOrigin` + (optionally) loopback-RemoteAddress check.
2. A reusable `BridgeServer` class with the ephemeral-port + shared-singleton + refcount + send-timeout + pending-map drain pattern that takes an `IBridgeHandler` interface for per-agent semantics.

This is out of scope for the current PR (would also touch all the already-merged agents), but I'll file a tracking issue so it doesn't get lost. cc @TalZaccai.

dotnet/visualStudioTypeAgent/Bridge/BridgeDiscovery.cs
TalZaccai: Better fix in 51978540ec2c3f2400d754144ced56c4c5421dbe: moved the VSIX project from `ts/packages/agents/visualStudio/host/csharp/` to `dotnet/visualStudioTypeAgent/`. The repo already has `dotnet/.editorconfig` with `csharp_prefer_braces = true:error` and a 350-rule StyleCop-ish ruleset, but the VS extension was the only C# project under `ts/` and `ts/.editorconfig` declares `root = true` with zero C# rules. The project is now under `dotnet/` and inherits that policy automatically ΓÇö `IDE0011` and friends will now light up in the IDE / build. The earlier brace-cleanup commit (b8b9e8ded) is still valid prep work for the rule.

PR #2429, opened 2026-06-03
chore(graph-utils): export GraphClient base class

TalZaccai:
Make the `GraphClient` base class available alongside `CalendarClient` / `MailClient` (which already extend it) so downstream agents can build new Microsoft Graph–backed clients without copying the auth/transport plumbing.

### Change
`ts/packages/agents/agentUtils/graphUtils/src/index.ts` — add `GraphClient` to the existing named re-export from `./graphClient.js`.

### Risk
Additive-only. No existing consumer breaks because the export was previously narrower; all current callers use the more specific subclasses or `ErrorResponse`.

### Test
- `pnpm --filter graph-utils build` ✓
- `pnpm --filter graph-utils prettier` ✓
- Verified `GraphClient` shows up in `dist/index.d.ts` after the change.

### DIFF
--- a/ts/packages/agents/agentUtils/graphUtils/src/index.ts
+++ b/ts/packages/agents/agentUtils/graphUtils/src/index.ts
@@ -14,7 +14,7 @@ export {
 export { createCalendarGraphClient, CalendarClient } from "./calendarClient.js";
 export { createMailGraphClient, MailClient } from "./mailClient.js";
 export { GraphEntity } from "./graphEntity.js";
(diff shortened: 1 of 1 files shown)


TalZaccai:
Superseded by a combined PR that adds the actual motivation for exporting GraphClient: a GraphClientOptions parameter so subclasses can target a different AAD app (custom env-var prefix, persisted auth-record filename, token-cache name, and OAuth scopes). Re-opening as a single PR from `talzacc/graph-client-options-refactor` since the export alone has no in-tree consumer.

PR #2498, opened 2026-06-16
Address open CodeQL shell-command alerts (#262, #316-319)

TalZaccai:
## Summary

Addresses the 5 open medium-severity CodeQL code-scanning alerts on `main`. All are real fixes — no suppressions or overrides to maintain.

| Alerts | Rule | File(s) | Fix |
|---|---|---|---|
| [#316](https://github.com/microsoft/TypeAgent/security/code-scanning/316)-[#319](https://github.com/microsoft/TypeAgent/security/code-scanning/319) | `js/shell-command-constructed-from-input` | `actionGrammar/src/generation/cliPath.ts`, `agentSdk/src/node/cliPath.ts` | `execSync(\where/which \\)` → `execFileSync(lookup, [name])` |
| [#262](https://github.com/microsoft/TypeAgent/security/code-scanning/262) | `js/shell-command-injection-from-environment` | `agentServer/client/src/agentServerClient.ts` | `cmd /c start /B node` → direct `node` spawn with `detached` + `windowsHide` |

## Notes

**#316–#319** — `resolveCliOnPath(name)` built a shell string from its argument. Switched to `execFileSync` with `name` as a separate argv entry, so it runs no shell and the value can never be interpreted as a command.

**#262** — A prior PR (#2428) fixed only the *non-hidden* (pwsh) branch; CodeQL then re-flagged the *hidden* branch, which routed the `__dirname`-derived `serverPath` through `cmd … start /B node`. Replaced with a direct `node` spawn — no shell at all. `detached: true` (DETACHED_PROCESS) keeps the server alive past the client, and `windowsHide: true` (CREATE_NO_WINDOW) avoids a console-window flash. This mirrors the existing `studioServiceLauncher` hidden-spawn pattern.

## Verification

- `src` builds pass for `@typeagent/agent-sdk`, `action-grammar`, and `@typeagent/agent-server-client`.
- Verified on Windows: `execFileSync` `where` resolution returns the expected path, and a `detached + windowsHide + stdio:ignore` node spawn runs hidden and to completion.


ts/packages/agentSdk/src/node/cliPath.ts
```diff
@@ -5,7 +5,7 @@
 // `node:child_process` dependency is isolated from the package's main entry,
 // which is also consumed by browser/renderer bundles.
 
-import { execSync } from "node:child_process";
+import { execFileSync } from "node:child_process";
```
robgruen: These cliPath.ts files look eerily similar...is there a common library that both of these projects use where we can put this file and then reuse it?

ts/packages/agentServer/client/src/agentServerClient.ts
```diff
@@ -396,19 +396,17 @@ function spawnAgentServer(
...
-                const child = spawn(
-                    "cmd.exe",
-                    ["/c", "start", "/B", "node", ...args],
-                    {
-                        stdio: "ignore",
-                        windowsHide: true,
-                    },
-                );
+                const child = spawn("node", args, {
+                    detached: true,
```
robgruen: I would actually make these parameters so the caller can control this behavior.

robgruen (approved):

ts/packages/agentSdk/src/node/cliPath.ts
TalZaccai: Done — hoisted the single implementation into `@typeagent/common-utils` (node-only export). agentSdk, actionGrammar, and a third copy in `kp` now re-export from it, so there's one source of truth. I removed common-utils's unused `@typeagent/agent-sdk` dependency so the graph stays acyclic (agent-sdk -> common-utils). Bonus: kp's copy still had the original `execSync` pattern, so it picks up the security fix too.

ts/packages/agentServer/client/src/agentServerClient.ts
TalZaccai: Done — added an optional `AgentServerSpawnOptions` ({ detached?, windowsHide?, stdio? }) threaded through `ensureAgentServer` -> `spawnAgentServer` so callers can control these for the direct-node spawn paths. Defaults preserve the current background-server behavior (detached + windowsHide + ignored stdio). The type is exported from the package barrel.

PR #2509, opened 2026-06-18
TypeAgent Studio — real replay path + Impact Report

TalZaccai:
Replaces the "find a regression" flow's placeholder identity comparison with a
real replay path, ships the Impact Report webview, and hardens the service
connection with a shared liveness heartbeat.

## Replay engine (`typeagent-core`)

- **Grammar replay** — utterances are matched against the agent's compiled
  grammar (the real deterministic, no-LLM path). Ships `builtInEntities.agr` in
  the bundle so it works when packaged.
- **Schema-enriched matching** — projects the agent's action-schema metadata
  (checked wildcards, params) onto the grammar and matches through the real
  `GrammarStore`, falling back to plain grammar matching when the schema can't be
  discovered.
- **Construction-cache consult** — the live working-tree side consults the
  agent's real per-session construction cache before the grammar, mirroring the
  dispatcher's cache-first path. It is hash-gated exactly as the dispatcher gates
  it, so a schema edit invalidates the cached entries (a cache hit is reported as
  `hit`, a grammar fall-through as `miss`). Consulted for the working tree only
  (caches are runtime artifacts, never committed or read at a git ref) and
  degrades cleanly when the cache is missing or stale — which is itself the
  regression signal. Discovered under the user data dir, overridable via
  `TYPEAGENT_STUDIO_CONSTRUCTION_CACHE` for determinism.

## Impact Report + `webviewKit` (`typeagent-studio`)

- Minimal reusable `webviewKit`: strict CSP/nonce HTML host, singleton-panel
  manager, typed host↔webview protocol, browser-neutral view model.
- Impact Report webview drives the replay over the service channel and renders
  the comparison, with a context header, A/B version controls, durable state
  across navigate-away/reload, and run-error surfacing.
- **Per-side method labels** — since the construction cache is live-only, a git
  ref can never consult it. Each A/B field shows how that side actually resolved,
  and cache-served hits are tagged (`B:hit·cache` vs grammar `B:miss·grammar`),
  so the run-level label never implies the cache served a side that didn't read
  it.

## Connection-aware UX

- Corpora / Event Log / Collisions views show "Connect to Studio service" welcome
  content when the service is down (mirroring Sandboxes).
- Corpora view refreshes after any corpus change made through the UI; externally
  hand-edited corpus files are picked up by the manual refresh (no always-on file
  watcher). Clearer seed-corpus affordance; Impact Report reachable from the
  Corpora view title.
- Fixes the Impact Report double-render on navigate-away/back.

## Shared WebSocket liveness heartbeat (`utils`)

- Extracts a client-side ping/pong watchdog alongside the existing server-side
  one into `@typeagent/websocket-utils`; `websocket-channel-server` re-exports
  both for back-compat (other agents untouched).
- The Studio service client uses it to detect a silently-dropped service (a
  half-open socket that never emits `close`) and drive the existing reconnect
  path — fixing a stale "connected" status. The watchdog is re-armed on every
  reconnect, and the backoff guards against an empty interval list collapsing
  into a tight loop.


ts/packages/typeagent-studio/src/extension.ts
```diff
@@ -501,6 +501,28 @@ export function activate(context: vscode.ExtensionContext): void {
...
+                "corpus/*.utterances.jsonl",
+            ),
+        );
+        const refreshOnCorpusChange = () => corpusTree.refresh();
+        context.subscriptions.push(
+            corpusWatcher,
+            corpusWatcher.onDidCreate(refreshOnCorpusChange),
+            corpusWatcher.onDidChange(refreshOnCorpusChange),
+            corpusWatcher.onDidDelete(refreshOnCorpusChange),
+        );
```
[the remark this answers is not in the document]
TalZaccai: Fixed in e7f41d99d. Removed the always-on corpus FileSystemWatcher. The two in-UI corpus-create paths (seedInRepoCorpus, ddExternalCorpus) already call corpusTree.refresh(), and a manual refresh remains for externally hand-edited files — so behavior now matches the description with no idle file-watching.

ts/packages/typeagent-studio/src/studioServiceClient.ts
```diff
@@ -80,6 +90,9 @@ export class StudioServiceClient {
         if (options.onClose) {
             socket.on("close", options.onClose);
         }
+        attachClientHeartbeat(socket, {
+            intervalMs: options.heartbeatMs ?? 10_000,
+        });
```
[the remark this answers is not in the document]
TalZaccai: Fixed in e7f41d99d. 
econnect() now re-attaches ttachClientHeartbeat(socket, { intervalMs: this.heartbeatMs }) — the interval is computed once in connect() and stored on the client, so every reconnected socket gets the liveness ping. Also guarded an empty backoff schedule so reconnect can't tight-loop.

ts/packages/typeagent-studio/package.json
```diff
@@ -22,7 +22,7 @@
     "clean": "rimraf --glob dist dist-pub *.vsix",
     "compile": "node esbuild.mjs",
     "deploy:local": "npm run package && code --install-extension dist-pub/typeagent-studio.vsix --force",
-    "package": "mkdirp dist-pub && vsce package --allow-star-activation --allow-missing-repository --no-dependencies --allow-package-secrets npm -o dist-pub/typeagent-studio.vsix",
+    "package": "mkdirp dist-pub && vsce package --allow-star-activation --allow-missing-repository --no-dependencies --allow-package-secrets npm --allow-package-secrets privatekey -o dist-pub/typeagent-studio.vsix",
```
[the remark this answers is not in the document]
TalZaccai: Kept intentionally — this is a concrete, unavoidable false positive. The esbuild bundle pulls in a crypto/JWT dependency whose key-serialization code contains literal PEM delimiters:

```
-----BEGIN PRIVATE KEY-----
${n.toString("base64")}
-----END PRIVATE KEY-----
```

That is library code wrapping a runtime-supplied key (the surrounding code is JWT alg handling — RS256/ES256/etc. and `createCipheriv`); no actual key material is embedded in the package. vsce's secret scanner matches the `-----BEGIN/END PRIVATE KEY-----` markers, and the `--allow-package-secrets` opt-out is only available per secret *type*, not per file, so it can't be scoped more tightly. Without the flag `vsce package` fails. The `npm` allow-list is there for the same reason on the registry-token scanner.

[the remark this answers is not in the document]
TalZaccai: Update: rather than keep the suppression, I removed `--allow-package-secrets privatekey` entirely (7e6662179) by getting both PEM sources out of the package:

1. **mongodb (real code in the bundle).** The PEM came from the `mongodb` driver's client-side-encryption crypto callbacks, pulled in transitively because `@typeagent/telemetry`'s barrel re-exports `createMongoDBLoggerSink`. Studio/service/core never call that sink, so the driver was dead weight — I aliased `mongodb` to an inert stub in the studio esbuild config, dropping the driver and its PEM from `extension.js`/`studio-service.js`.
2. **@azure/msal-node (PEM example in a source comment).** This only reached the package via the 40 MB sourcemaps. `.vscodeignore` had `*.map`, which only matches root-level files, so `dist/**/*.map` shipped anyway — widened to `**/*.map`.

`vsce package` now succeeds with only `--allow-package-secrets npm` (a bundled `npm i --save-dev @types/...` doc string that trips the npm-token scanner — a genuine, narrow false positive), and the `.vsix` drops from ~45 MB to 2.5 MB. Private-key scanning is fully re-enabled.

ts/packages/typeagent-core/test/fixtures/playerConstructions.json
```diff
@@ -0,0 +1,363 @@
+{
```
robgruen: Don't we have a version of this type of file along side the construction code?  If so can we reference that one instead of this one.  I worry about drift...

ts/packages/typeagent-studio/src/eventLogTreeProvider.ts
```diff
@@ -40,11 +40,21 @@ export class EventLogTreeProvider
...
     }
 
+    /** Reflect the studio service connection state (drives the empty view). */
+    setConnected(connected: boolean): void {
+        if (connected === this.connected) {
+            return;
+        }
+        this.connected = connected;
+        this.refresh();
+    }
```
robgruen, line 56: these "treeProvider" classes have some shared code...could we condense and inherit from a generic treeProvider class?

ts/packages/typeagent-studio/src/studioServiceConnection.ts
```diff
@@ -49,7 +49,14 @@ export class StudioServiceConnection {
             token?: string;
         } = {},
     ) {
-        this.backoffMs = options.backoffMs ?? [2000, 4000, 8000, 15000];
+        const DEFAULT_BACKOFF = [2000, 4000, 8000, 15000];
```
robgruen, line 52: why have fixed backoff options? Would it be better to have some sort of exponential with some sort of limit/restart option?

robgruen (approved):

ts/packages/typeagent-studio/src/studioServiceConnection.ts
TalZaccai: Good call - fixed in a follow-up PR. I pulled the reconnect delay out into a shared `createBackoff({ baseMs, maxMs })` helper in `@typeagent/websocket-utils` that does exponential backoff (doubling) with a cap, plus reset-on-success. Both this connection (2s->15s) and the vscode-shell agent-server bridge (2s->30s, previously a linear ramp) now use the one strategy instead of hand-rolling their own.

ts/packages/typeagent-studio/src/eventLogTreeProvider.ts
TalZaccai: Agreed - done in a follow-up PR. The four tree providers (Event Log, Collisions, Sandboxes, Corpora) now extend a shared `BaseStudioTreeProvider<TNode>` that owns the change emitter, the connected flag + `setConnected` gate, `refresh`, and the descriptor->TreeItem field mapping. Each subclass keeps only what genuinely differs (`getChildren`, collapsible state, icon/command decoration). I also ran a broader intra-package dup scan off the back of this and folded in a shared collapse-and-truncate text helper (it had four hand-rolled copies).

ts/packages/typeagent-core/test/fixtures/playerConstructions.json
TalZaccai: Good catch. Done in this PR (98251a7b0): the spec now points at the shipped player cache that lives alongside the construction code (defaultAgentProvider/data/explainer/v5/constructions.json), and I deleted the duplicate fixture. The two had the identical namespace/hash, so the hand-maintained copy was purely a drift risk. Verified the resolver tests still pass against the canonical file: match('pause') -> player.pause and match('play despacito') -> player.playArtist.

PR #2610, opened 2026-07-07
Fix open CodeQL code-scanning alerts

TalZaccai:
Resolves the 6 open CodeQL code-scanning alerts.

## Code fixes

- **`displayRender.ts` — `js/incomplete-multi-character-sanitization` (#321):** `htmlToText` stripped tags with a single `replace(/<[^>]*>/g, "")` pass, which nested/overlapping tags (e.g. `<scr<script>ipt>`) can survive by reconstructing a tag. Now strips iteratively until the string is stable.
- **`displayRender.ts` — `js/incomplete-sanitization` (#322):** `escapeTableCell` escaped `|` and newlines but not backslashes, so a trailing `\` could escape the escaping. Now escapes `\` first.
- **`build-msi.mjs` — `js/incomplete-sanitization` (#323):** removed the unused `quoteCmdArg` helper (dead code) that escaped quotes but not backslashes.
- **`agentServerClient.ts` — `js/shell-command-constructed-from-input` (#327):** the visible-window launch concatenated `serverPath` into a PowerShell `-Command` string routed through `cmd.exe /c start`. Now `serverPath` is passed via the child's environment and the command is handed to PowerShell as a base64 `-EncodedCommand`, so no path text reaches either shell parser. Behavior (new window, `-NoExit`) is unchanged.

## Dismissed as false positive

- **#324 `sign-msi.mjs` (clear-text-logging):** logs the *name* of the KeyVault secret to help operators grant `az` access, not the password value.
- **#325 `typeagent-serve.mjs` (shell-injection-from-environment):** the flagged input is `process.execPath` (the Node binary), not attacker-controlled; Node also quotes array args to `cmd.exe`.

### DIFF
--- a/ts/packages/agentServer/client/src/agentServerClient.ts
+++ b/ts/packages/agentServer/client/src/agentServerClient.ts
@@ -556,10 +556,11 @@ function spawnAgentServer(
                 const psExe = fs.existsSync(pwsh7) ? pwsh7 : "powershell.exe";
                 // Pass serverPath to the child through the environment rather
                 // than the command line so it never reaches cmd.exe or the
-                // PowerShell parser as text. The command itself is a constant
-                // that reads $env:TYPEAGENT_SERVER_PATH, and it is handed to
-                // PowerShell as a base64 -EncodedCommand blob, whose alphabet
-                // is inert to both cmd.exe and PowerShell.
+                // PowerShell parser as text. Only the numeric port and
+                // idleTimeout are interpolated into the command; the path is
+                // read from $env:TYPEAGENT_SERVER_PATH at runtime. The command
+                // is handed to PowerShell as a base64 -EncodedCommand blob,
+                // whose alphabet is inert to both cmd.exe and PowerShell.
                 const psCommand =
                     `& node $env:TYPEAGENT_SERVER_PATH --port ${port}` +
                     (idleTimeout > 0 ? ` --idle-timeout ${idleTimeout}` : "");
--- a/ts/packages/agentServer/client/src/agentServerClient.ts
+++ b/ts/packages/agentServer/client/src/agentServerClient.ts
@@ -503,6 +503,36 @@ export interface AgentServerSpawnOptions {
     stdio?: StdioOptions;
 }
 
+// Locate a PowerShell executable on Windows. Prefer PowerShell 7+ (pwsh.exe)
+// from its standard install locations or PATH, so installs outside the default
+// "C:\Program Files\PowerShell\7" directory (winget, x86, a relocated Program
+// Files) are still found. Fall back to Windows PowerShell (powershell.exe),
+// which is always present. An absolute path is returned when one is found so
+// the result does not depend on the launcher's PATH.
+function resolvePowerShellExe(): string {
+    const candidates: string[] = [];
+    for (const base of [
+        process.env["ProgramFiles"],
+        process.env["ProgramW6432"],
+        process.env["ProgramFiles(x86)"],
+    ]) {
+        if (base) {
+            candidates.push(path.join(base, "PowerShell", "7", "pwsh.exe"));
+        }
+    }
+    for (const dir of (process.env.PATH ?? "").split(path.delimiter)) {
+        if (dir) {
+            candidates.push(path.join(dir, "pwsh.exe"));
+        }
+    }
+    for (const candidate of candidates) {
+        if (fs.existsSync(candidate)) {
+            return candidate;
+        }
+    }
+    return "powershell.exe";
+}
+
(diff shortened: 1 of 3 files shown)


ts/packages/agentServer/client/src/agentServerClient.ts
```diff
@@ -554,14 +554,20 @@ function spawnAgentServer(
...
-                const psQuote = (s: string) =>
-                    "'" + s.replace(/'/g, "''") + "'";
-                const psCommand = `& node ${psQuote(serverPath)} --port ${port}${idleTimeout > 0 ? ` --idle-timeout ${idleTimeout}` : ""}`;
-                const psArgs = ["-NoExit", "-Command", psCommand];
+                // Pass serverPath to the child through the environment rather
+                // than the command line so it never reaches cmd.exe or the
+                // PowerShell parser as text. The command itself is a constant
+                // that reads $env:TYPEAGENT_SERVER_PATH, and it is handed to
+                // PowerShell as a base64 -EncodedCommand blob, whose alphabet
+                // is inert to both cmd.exe and PowerShell.
```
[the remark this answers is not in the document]
TalZaccai: Good catch - reworded the comment to say only the numeric port/idleTimeout are interpolated and the path is read from \ at runtime. Fixed in 4bc1b9632.

ts/packages/agentServer/client/src/agentServerClient.ts
```diff
@@ -554,14 +554,21 @@ function spawnAgentServer(
             } else {
                 const pwsh7 = "C:\\Program Files\\PowerShell\\7\\pwsh.exe";
```
robgruen: This can fail if this path is wrong or different.  Is there a more robust way to get this path?

ts/packages/vscode-chat/src/displayRender.ts
```diff
@@ -150,17 +153,27 @@ function stripMarkdownHtml(text: string): string {
...
-            html
-                .replace(/<\s*br\s*\/?\s*>/gi, "\n")
-                .replace(/<\s*\/\s*(p|div|tr|li|h[1-6])\s*>/gi, "\n")
-                .replace(/<\s*\/\s*(td|th)\s*>/gi, " | ")
-                .replace(/<[^>]*>/g, "")
-                .replace(/[ \t]+\n/g, "\n")
-                .replace(/\n{3,}/g, "\n\n"),
-        ),
-    ).trim();
+    const withStructure = html
```
robgruen, line 156: We have a dependency on DOMPurify...can it do this for us?  I'd prefer to have one solution or the other (if we can) and not mixed HTML scrubbing methods.

robgruen (approved):

ts/packages/agentServer/client/src/agentServerClient.ts
TalZaccai: Good call - hardened this in e4af9aa8c. It now searches the standard PowerShell 7 install locations (ProgramFiles / ProgramW6432 / ProgramFiles(x86)) plus PATH for pwsh.exe, and falls back to powershell.exe (always present on Windows) if none are found - so a non-default PS7 install (winget, x86, relocated Program Files) still gets picked up. FWIW this line predates this PR (came in with #2199); it wasn't part of the CodeQL fix, but worth tightening while I was in here.

ts/packages/vscode-chat/src/displayRender.ts
TalZaccai: vscode-chat doesn't actually depend on DOMPurify (shell and chat-ui do, but not this package), and it runs in the VS Code extension host - plain Node with no DOM - so DOMPurify would need jsdom brought in as a new dependency to work here. On top of that, this path (htmlToText) converts HTML to plain text for renderDisplayToText, whereas DOMPurify sanitizes HTML and hands back HTML, so it isn't a drop-in for what this function does. This PR didn't introduce a new scrubbing method - it just hardened the existing regex htmlToText (made the tag strip iterative) to close the CodeQL alert. If we want to standardize on a single HTML approach with a real parser in this package, I'm happy to do that as a follow-up, but I'd keep it out of this security-fix PR.

ts/packages/agentServer/client/src/agentServerClient.ts
TalZaccai: Refined this in 987138189 after a review pass: it still detects PS7 across the standard locations + PATH, but now hands start a bare name (pwsh.exe / powershell.exe) rather than the resolved absolute path, so no environment-derived path flows through cmd.exe. start resolves the name via PATH/App Paths, which every standard PS7 install registers.

PR #2618, opened 2026-07-07
Author schema keyword files and backfill agents

GeorgeNgMsft:
# Per-Agent Keyword-Vector Authoring for Context-Weighted Collision Resolution

## Summary
Builds on the contextSelector collision-resolution tier (V1, #2580) by giving every
agent action a **committed keyword vector** — the topical "fingerprint" the
contextSelector weighs to break grammar/cache collisions. Each schema now ships a
`<schema>.keywords.json` sibling ("Source 1"), authored once via LLM distillation
(with a deterministic lexical floor as fallback), which the read path prefers over
live lexical extraction.

## What's included
- **Keyword file format + read path** — per-agent `<schema>.keywords.json` stored
  beside the schema source; `KeywordIndex` prefers it, falling back to the lexical
  floor when absent. Tokens are canonicalized through the shared tokenizer so
  committed vectors always match at scoring time.
- **Producers** — an LLM distiller (`keywordDistiller`) preferred, backed by the
  existing lexical extractor as a guaranteed floor (`keywordProducer`).
- **`@collision keywords backfill [--llm] [--force] [schema…]`** — generates /
  refreshes the committed files across the whole roster; groups by source file so
  co-located schemas merge rather than overwrite; skips inline/dynamic agents.
- **`@config collision contextSelector`** — tune the abstain `margin` and related
  thresholds at runtime.
- **Backfilled vectors** — LLM-distilled keyword files committed for all built-in
  agents (Office agents committed separately in their own repo).

## Validation
- Full dispatcher suite green (1073 tests); build clean; complexity ratchet passes.
- End-to-end check: a committed synonym present only in the keyword file flips a
  scorer decision the lexical floor cannot make.

## Notes
- Merged latest `main` (incl. #2580) — the branch is current with `origin/main`.
- Keyword files are committed source (the LLM producer isn't a deterministic build
  step), stored per-agent to match how `.pas.json` / `.ag.json` artifacts live.
- Tested manually with primed conversations
image
(words only in the first generation of the keyword file)
image
image
- Also tested with a validation script that instantiates and calls the relevant classes / functions (with context keywords unique to the file and not the in-memory dynamic keywords) such as keywordIndex, tfIdfScorer, action-schema parser, etc. Result : 
image


robgruen:
do we have thoughts/ideas/plan for localization for keywords?

also, what's the evaluation plan to determine if this helps or not?

ts/packages/dispatcher/dispatcher/src/context/contextSelector/keywordDistiller.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+function buildPrompt(input: KeywordExtractionInput, topN: number): string {
+    const actionDocs = stripExampleValues(
+        (input.actionComments ?? []).join(" "),
+    );
+    const params = (input.paramNames ?? []).join(", ").trim();
+    const paramDocs = stripExampleValues((input.paramComments ?? []).join(" "));
+    const schemaDesc = stripExampleValues(input.schemaDescription ?? "");
+    return [
```
robgruen, line 63: Why not use TypeChat here?

ts/packages/dispatcher/dispatcher/README.md
```diff
@@ -344,19 +344,20 @@ Each event carries `kind` (detection point), `strategy`, `candidates[]` (with pe
...
-| `@config log db [on\|off]`                                         | Toggle DocumentDB upload (gates remote sink — independent of the per-session local capture).                                                                             |
+| Command                                                            | Effect                                                                                                                                                                                           |
+| ------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `@config collision`                                                | Render the current collision config as an HTML status table.                                                                                                                                     |
```
robgruen: Can you update the command reference with these new options: https://github.com/microsoft/TypeAgent/blob/main/ts/docs/overview/command-reference.md

robgruen (approved):

robgruen (approved):

Issue #2499, opened 2026-06-16
NFA compiler rejects grammar-optimizer output for grammars with a shared optional prefix


GeorgeNgMsft:
**Area:** `@typeagent/action-grammar` (NFA compiler + grammar optimizer), `@typeagent/action-grammar-compiler` (`agc`)
**Labels:** bug, action-grammar, nfa, grammar-optimizer
**Status:** root-caused; a per-agent stopgap is already shipped (see *Current mitigation*). This issue tracks the proper core fix.

---

## Summary

When `agent-server` runs the **NFA grammar system** (the default — `grammarSystem: "nfa"`), any grammar whose top-level alternatives all begin with the **same optional sub-rule** fails to compile to an NFA. The default `agc` optimizer rewrites those alternatives into a shape the NFA compiler refuses, so the schema silently loses its grammar fast-path and falls back to LLM translation. One schema (`excel.excel-flow`) surfaces it as a hard `console.error` on every startup; the rest fail quietly.

The underlying problem is an **inconsistency between the two matchers in the package**: the AST-walking (completion-based) matcher accepts and correctly evaluates the optimizer's output, but the NFA compiler does not.

---

## Bug, or design gap? (triage framing)

Both — and they map cleanly onto the two blockers below:

- **The `tailCall` incompatibility is a design choice that's no longer sufficient.** `tailFactoring` /
  `promoteTailRulesParts` emit matcher-only constructs *by design* (documented at
  `grammarOptimizer.ts:203-256`). The original assumption was that NFA consumers would not use the
  optimized output. That assumption was broken by the **pipeline + defaults**: a single `.ag.json` is
  built once by `agc` (optimized) and consumed by whichever matcher is active, and `agent-server` now
  defaults to `grammarSystem: "nfa"`. The seam (one artifact, two matchers with different capabilities)
  was never updated when NFA became the default. → addressed by **Option B**.
- **The value-less multi-term rejection is a genuine bug.** `factorCommonPrefixes` is *not* documented as
  matcher-only; the AST matcher accepts and correctly evaluates its output; and the NFA compiler already
  has the mechanism to handle it (`getImplicitDefaultValue` / its single-variable derivation) but applies
  it inconsistently and throws first. The optimizer and NFA compiler disagree about a shape that is
  supposed to be valid. → addressed by **Option A**.

This is why the recommended direction is **A + B together**: fix the inconsistency *and* update the
build seam that NFA-as-default outgrew.

---

## How to reproduce

[...]

PR #2629, opened 2026-07-08
Fix NFA compilation failure for grammars with shared optional prefixes

GeorgeNgMsft:
See issue https://github.com/microsoft/TypeAgent/issues/2499

**Problem**

The NFA compiler rejected grammars where multiple top-level alternatives share
a common optional prefix (e.g.  <Start> = <a> | <b>; <a> = <P> foo -> ...; <P> = (please)?; ). The optimizer's  factorCommonPrefixes  pass can produce a
value-less, multi-term rule in this shape, which the NFA compiler had no
general way to derive a value for — it only special-cased a narrower "single
wildcard/number part" shape.

**Root-cause fix**

•  nfaCompiler.ts  now derives a rule's effective value (explicit  -> 
expression, or implicit single-variable-part forwarding) using the same
general logic the AST-based optimizer already used, instead of its old,
narrower special case.
• Extracted this rule-value-derivation logic into a new shared module,
 grammarValueDeriver.ts  — a single source of truth now used by both the
grammar optimizer and the NFA compiler (previously each had its own
slightly different implementation).
• Side effect of consolidation: nested rules with a single bound variable and
no explicit value now correctly forward that value in the NFA path instead
of silently discarding it.
• Harmonized error messages for rules with no value and no single
variable-bearing part to forward, across both the "ambiguous" and
"missing" throw paths.

**Follow-up hardening**

• Hoisted the separate  tailCall  (NYI tail-RulesPart) rejection check to a
single, earliest, unconditional location ( normalizeRule ), run before any
rule rewriting — closing a gap where one rule shape (a single-part
tailCall rule) could slip past the check and surface a confusing,
differently-worded error later in compilation.
• Unified the optimizer's  getImplicitDefaultValue  onto the same shared
 deriveEffectiveValue , removing a second, redundant implementation.

**Testing**

• Added targeted unit tests for the shared derivation helper and regression
tests for previously-unreachable error paths (0-part / 1-part rules with
no value, and the single-part tailCall shape).
• Cleaned up stale comments/tests referencing removed or relocated
functions; verified no redundant test coverage was introduced.
• Full suite: 75/75 suites, 16153/16155 tests passing (2 skipped), clean
build, prettier clean.


ts/packages/actionGrammarCompiler/src/commands/compile.ts
```diff
@@ -86,7 +86,9 @@ export default class Compile extends Command {
                 : {
                       startValueRequired: true,
                       schemaLoader,
-                      optimizations: recommendedOptimizations,
+                      // NFA is agent-server's default grammar system;
```
curtisman: NFA shouldn't be the default yet. recommended optimization should be made to be supported instead of getting it disabled.

ts/packages/actionGrammar/test/nfaFactoredPrefixValues.spec.ts
```diff
@@ -0,0 +1,107 @@
+// Copyright (c) Microsoft Corporation.
```
curtisman, line 1: There might be existing grammarMatcher* test covering these.   
you need to set `TYPEAGENT_TEST_VARIANTS=nfa,dfa` to turn those on.  Currently, it has 116 failure.
If it is not covered, you can add the test in to using describeForEachMatcher, so that the test case is ran for all AST, NFA and DFA.

Make sure you change doesn't have new failures.

ts/packages/actionGrammar/src/nfaCompiler.ts
```diff
@@ -473,6 +473,26 @@ function isSingleVariableRule(rule: GrammarRule): { variable: string } | false {
...
 }
 
+/**
+ * For a value-less multi-part rule, find its single variable-bearing
+ * part so that part's value can be forwarded as the rule's effective
+ * value (mirrors `getImplicitDefaultValue` in grammarOptimizer.ts).
+ * Returns `"ambiguous"` if more than one part carries a variable, or
+ * `undefined` if none do.
+ */
+function findSingleVariableBearingPart(
```
curtisman: This (and the singleVar case in the caller) look like a duplicate of `getImplicitDefaultValue` in `grammarOptimizer.ts`.  
Can we reuse or extract into a helper function?

ts/packages/actionGrammar/test/nfaFactoredPrefixValues.spec.ts
GeorgeNgMsft: image

Also switched the test to use describeForEachmatcher

ts/packages/actionGrammar/src/nfaCompiler.ts
GeorgeNgMsft: Extracted it from both grammarOptimizer and nfaCompiler into a shared function to be used in both.

ts/packages/actionGrammarCompiler/src/commands/compile.ts
GeorgeNgMsft: Reverted the default change and added a opt-in flag "--nfa-safe" to gaurantee compiling to NFA safely and preserve some optimizations

ts/packages/actionGrammar/test/nfaFactoredPrefixValues.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+        });
+
+        expect(() =>
+            compileGrammarToNFA(grammar, "factored-prefix-tail"),
+        ).toThrow(/tail/i);
+    });
+
+    it("evaluates an explicit -> value expression through a manually-factored shape", () => {
+        const agr2 = `
+<Start> = <a> | <b>;
```
curtisman: are these test not covered in existing tests?  Just don't want to have duplicate coverage.

Probably want to create an issue to change the matcher/completion test harness so that to only disable NFA/DFA if they are failing right now.

ts/packages/actionGrammarCompiler/src/commands/compile.ts
```diff
@@ -68,6 +69,11 @@ export default class Compile extends Command {
                 "Disable grammar optimizations (produces an unoptimized AST that preserves the 1:1 correspondence between top-level rules and the original source — useful for diagnostics).",
             default: false,
         }),
+        "nfa-safe": Flags.boolean({
```
curtisman: I kinda prefer not to introduce this flag because it indicate that this is the long term picture, instead of transient NYI.  But if it is useful right now, make a TODO to make sure we get rid of it once NFA has paritity.

ts/packages/actionGrammar/src/nfaCompiler.ts
```diff
@@ -473,6 +475,58 @@ function isSingleVariableRule(rule: GrammarRule): { variable: string } | false {
...
+    }
+    if (rule.parts.length <= 1) {
+        return undefined;
+    }
+    const singleVarPart = findSingleValueBearingPart(rule.parts);
+    if (singleVarPart === "ambiguous") {
+        throw new Error(
+            `${describeRule()} has ${rule.parts.length} terms but no value expression, ` +
+                `and more than one part carries a variable - the implicit value is ambiguous. ` +
+                `Multi-term rules must have an explicit value expression (using ->).`,
```
curtisman: From my recollection, not all rule needs to have a value.   It is only an error if a rule without a value is assign to a value.

So this is not true?

curtisman: https://github.com/microsoft/TypeAgent/blob/main/ts/docs/architecture/core/actionGrammar.md#implicit-value-behavior

ts/packages/actionGrammar/test/nfaFactoredPrefixValues.spec.ts
GeorgeNgMsft: Made a redundancy scan and cleaned up a couple of tests, thanks!

ts/packages/actionGrammarCompiler/src/commands/compile.ts
GeorgeNgMsft: This allows the nfa path to be used with excel agent (and any other similar scenario that triggers the problematic path) right now while keeping the optimizations, so we do need it. I have added a TODO for removal as suggested

ts/packages/actionGrammar/src/nfaCompiler.ts
GeorgeNgMsft: Thanks for catching that, I misunderstood the logic there. Now, only topic-level rules that have requireValue=true will error here when it is ambiguous. Optional rules will return undefined (like before) now as I have moved the check above

curtisman: We should error if singleVarPart is "undefined" too?
Also do we need to distinguish between ambiguous and "undefined".  This is just an "internal error" as the compiler should have caught this case.

ts/packages/actionGrammar/test/nfaFactoredPrefixValues.spec.ts
curtisman: Let's name this file `grammarMatcher<Name>`

ts/packages/actionGrammar/src/nfaCompiler.ts
```diff
@@ -451,26 +453,56 @@ function stripDispatch(part: RulesPart): RulesPart {
...
+            : `has ${rule.parts.length} terms`;
+    if (result.kind === "ambiguous") {
+        throw new Error(
+            `${describeRule()} ${termsDescription} but no value expression, ` +
+                `and more than one part carries a variable - the implicit value is ambiguous. ` +
+                `Rules must have an explicit value expression (using ->) unless exactly one part carries a variable.`,
+        );
     }
-    return false;
+    const hasTailCall = rule.parts.some(
```
curtisman: Make sure that tail call is noted as a NYI TODO, not a fundamental limitation.  

Also, shouldn't tail call always error for NFA/DFA path? It nay make more sense if we hoist that check somewhere else instead of mixing it with "no value" checking, which conflates the two continuet.

If you do that, then we can just use the original getImplicitValue to get CompileValueNode and undefined and error on undefined if required value.

GeorgeNgMsft: Good call on the NYI and you're right about the tailcall always being an error path. I've moved it into normalizeRules in nfaCompiler to error early. 

For getImplicitValue, I forgot that I added the error throwing to deriveEffectiveValue, I thought there was some earned difference there that we really wanted to throw that error for. I've unified them now

ts/packages/actionGrammar/src/grammarValueDeriver.ts
```diff
@@ -0,0 +1,119 @@
...
+    | { kind: "value"; value: CompiledValueNode }
+    | { kind: "ambiguous" }
+    | { kind: "none" };
+
+/**
+ * Compute a rule's effective value expression: its explicit `->` value
+ * when present, otherwise the implicit value it would produce (via
+ * `findSingleValueBearingPart`).
+ */
+export function deriveValue(rule: GrammarRule): ImplicitValueResult {
```
curtisman: I don't think there are other caller other then deriveEffectiveValue (other than test)?  This can be simplified if you merge it with deriveEffectiveValue.

ts/packages/actionGrammar/src/grammarOptimizer.ts
```diff
@@ -3241,11 +3239,15 @@ function trySubstituteMembers(
...
-                      parts: renamed.parts,
-                      value: renamed.value,
-                  });
+                : deriveEffectiveValue(
+                      {
+                          ...m,
+                          parts: renamed.parts,
+                          value: renamed.value,
+                      },
+                      () => "",
```
curtisman: switch this parameters and the next one, and make the call back optional

ts/packages/actionGrammar/src/nfaCompiler.ts
```diff
@@ -1528,15 +1516,11 @@ function compileRulesPartWithSlots(
...
-        let effectiveValue = rule.value;
-        if (!effectiveValue) {
-            const singleVar = isSingleVariableRule(rule);
-            if (singleVar) {
-                effectiveValue = { type: "variable", name: singleVar.variable };
-            }
-        }
+        const effectiveValue = deriveEffectiveValue(
+            rule,
+            () => "Nested grammar rule",
```
curtisman: Since you are already being precise here, you add the partId and the index in the alternatives array here.

ts/packages/actionGrammar/src/grammarOptimizer.ts
```diff
@@ -3241,11 +3239,15 @@ function trySubstituteMembers(
...
+                      false,
```
curtisman: also make this parameter default to false.

ts/packages/actionGrammar/src/nfaCompiler.ts
curtisman: Oh wait. requireValue is false here, so this is a useless parameter.  See comment in grammarOptimizer.ts about switching the order and defaulting the value.

ts/packages/actionGrammar/src/grammarValueTypeValidator.ts
```diff
@@ -498,15 +499,15 @@ export function classifyRuleValue(rule: GrammarRule): RuleValueKind {
     if (rule.value !== undefined) {
         return { kind: "explicit" };
     }
-    const variableParts = rule.parts.filter((p) => p.variable !== undefined);
-    if (variableParts.length === 1) {
+    const contributor = findSingleValueBearingPart(rule.parts);
```
curtisman: I feel like this is an exception for using a helper, since this is trying to fully interpret what the kind of value it is, it would be easier to read if just do a plain implementation.  Then I don't have to go and look at `findSingleValueBearingPart` so see what is going on?

curtisman (approved):

ts/packages/actionGrammar/src/grammarValueTypeValidator.ts
GeorgeNgMsft: As discussed offline, I have inlined the logic as the extracted function feels more like a code smell. Reverting that extraction as the prior logic was simple enough.

PR #2644, opened 2026-07-10
Add benchmarks and improvements for contextSelector

GeorgeNgMsft:
Strategy Comparison (Standalone Comparison 1000 tests)
image

Full Route Comparisons (250 dialogues)

image

image

image

Aggregate on normal traffic: the 88 collisions contextSelector confidently resolves (simple+realistic+hard) would each be an LLM round-trip under escalate-to-LLM. That's ~88 × 531 ms ≈ 47 s of serial LLM latency replaced by 88 × 8 µs ≈ 0.7 ms of local compute — same-or-better routing, zero regressions, effectively instant.

Note: the speed win is vs the LLM path only. Against first-match, both are cache-path (µs-scale) — there the win is accuracy (+29%), not latency. contextSelector adds only ~8 µs per collision it evaluates, far under the design's <5 ms cache-path budget.

Bottom line

• Normal traffic (150 scenarios): strictly additive vs both baselines — as-accurate-or-better routing with 0 regressions, while eliminating 88 LLM calls (~47 s → <1 ms).
• No-context: abstains 50/50 — never makes the no-signal case worse.
• Adversarial (50 scenarios): worse than the LLM (28% vs 56%); the negation guard shrank the gap (wrong-target 13→8, LLM-regressions 19→14) but sarcasm/quotes/dense ties need the LLM. Mitigated by shipping  detect: false  (off) and  abstainFallback: escalate-to-llm .

What & why: Adds a deterministic offline benchmark and comprehensive test coverage for the dispatcher's  contextSelector  collision-resolution tier, plus a negation-scope guard that stops loaded-negation phrases like "not the spreadsheet" from misrouting a collision to the rejected agent.

What's included

• Negation-scope guard (the only production behavior change) — the conversation signal now suppresses negated topic words before scoring, behind a  negationGuard  config lever (default on). Keyword extraction is unchanged, and the whole tier stays off by default ( detect: false ), so it's inert until enabled.
• Tier-1 tests (CI-gated, deterministic) — new context-extraction and scoring suites plus expanded tokenize/signal specs (127 contextSelector tests).
• Offline benchmark — a 250-scenario hand-authored dialogue corpus across 5 difficulty tiers replayed through the real scorer, an LLM-comparison arm, and a methodology write-up relating it to the static  @collision corpus  benchmark.

Results (measured)

• Adversarial tier: the guard cuts wrong-target misroutes 13 → 8 with zero safe-tier regressions.
• Normal traffic: contextSelector matches or beats both first-match (+29% routing accuracy) and the LLM with 0 regressions, while avoiding LLM calls


ts/packages/dispatcher/dispatcher/src/context/contextSelector/tokenize.ts
```diff
@@ -191,8 +191,53 @@ export type TokenizeOptions = {
...
+// cue words are stopwords, dropped before scoring, so the v1 extractor never saw
+// the negation and the negated words fired at full weight — the root cause of the
+// adversarial loaded-negation misroutes. When `dropNegatedSpans` is on, a scope
+// opens at a cue and content tokens inside it are suppressed. The scope closes at
+// the next clause boundary (punctuation), a reset connector ("but the config"),
+// or the end of the turn — so an idiom like "no problem, open the sheet" keeps
+// "sheet" (the comma closes the scope). Purely lexical and deterministic (§12);
+// contractions ("don't") are not handled in v1 — a parked lever. Off by default
+// so it never touches keyword vectors.
+const NEGATION_CUES: ReadonlySet<string> = new Set([
```
robgruen, line 220: I forsee problems here if the user is not speaking in English. We should note/plan localization...I occasionally speak German or Dutch with the thing to test it works very nicely.  It would be interesting to see how this affects that type of usage.

ts/packages/defaultAgentProvider/src/contextSelectorBench/compareLlm.mts
```diff
@@ -0,0 +1,344 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// contextSelector vs the full LLM resolution path. The offline benchmark scores
+// contextSelector against ground-truth labels; this harness answers a different
+// question the ship decision needs: how does contextSelector's routing compare
+// to what the STANDARD path would do with contextSelector OFF — i.e. the LLM
+// disambiguating the collision?
+//
+// Two arms, same 250 labeled collisions:
```
robgruen, line 10: Eventually we'd want to do this with 2500, 10000, or more phrases.

ts/packages/defaultAgentProvider/src/contextSelectorBench/compareLlm.mts
```diff
@@ -0,0 +1,344 @@
...
+import { loadConfigSync } from "../../../../../config/dist/index.js";
+import { openai } from "../../../../../aiclient/dist/index.js";
+import { loadRoster } from "./metricRoster.mjs";
+import { resolveReportPath, upsertLlmSection } from "./reportFile.mjs";
+import { SCENARIOS } from "./metricRealisticDialogue.mjs";
+import { RingBufferSignalSource } from "../../context/contextSelector/conversationSignal.js";
+import { TfIdfStrategy } from "../../context/contextSelector/strategy.js";
+import { DecisionConfig } from "../../context/contextSelector/decision.js";
+
+const HERE = path.dirname(fileURLToPath(import.meta.url));
```
robgruen: HERE seems a bit misrepresntative...why not something like THISFILE?

ts/packages/dispatcher/dispatcher/src/validation/contextselector/compareLlm.mts
```diff
@@ -0,0 +1,344 @@
...
+
+// Mirror production (session.ts negationGuard default on) so the CS-ON arm here
+// matches the guarded scorer the main metrics use. CS_NEGATION_GUARD=0 replays
+// the pre-guard baseline, consistent with metricRunner.
+const NEGATION_GUARD = process.env.CS_NEGATION_GUARD !== "0";
+
+// Concise, accurate one-liners for the featured agents so the LLM knows what
+// each does (a stand-in for the schema/action descriptions the real translator
+// sees).
+const DESC: Record<string, string> = {
```
robgruen: I can see drift here between the agent description and what's in this file.  Is there a reason we aren't using the description from the agent configs?

Also seems weird to have the benchmark live inside of the dispatcher.  Shouldn't this live outside of the dispatcher and the only thing that lives inside of it is the logic to load the context selector and do the switch?

ts/packages/defaultAgentProvider/src/contextSelectorBench/compareLlm.mts
```diff
@@ -0,0 +1,344 @@
...
+        "",
+        "Conversation:",
+        convo,
+        `Final request: "${ask}"`,
+        "",
+        "Candidate agents:",
+        `A) ${aSchema}: ${DESC[aSchema] ?? aSchema}`,
+        `B) ${bSchema}: ${DESC[bSchema] ?? bSchema}`,
+        "",
+        'Judge what the user ACTUALLY wants — account for negation ("not X"), sarcasm, and words that are quoting someone else. If the conversation genuinely does not favor one agent over the other, answer "unclear".',
```
robgruen: I can't tell where this is used from the GH review UI, but I recommend using TypeChat for the LLM call that makes this. Please disregard if you already are.

ts/packages/dispatcher/dispatcher/src/validation/contextselector/compareLlm.mts
```diff
@@ -0,0 +1,344 @@
...
+    const convo = dialogue.map((t, i) => `[${i + 1}] ${t}`).join("\n");
+    return [
+        "A user is talking to an assistant. Using the conversation so far, decide which of two agents should handle their final request.",
+        "",
+        "Conversation:",
+        convo,
+        `Final request: "${ask}"`,
+        "",
+        "Candidate agents:",
+        `A) ${aSchema}: ${DESC[aSchema] ?? aSchema}`,
```
robgruen: I'd also maybe consider adding a "neither" choice C...just in case!

robgruen (approved):

ts/packages/dispatcher/dispatcher/src/context/contextSelector/tokenize.ts
GeorgeNgMsft: I can definitely leave a note here, I realized that this was English specific but I didn't have a solution for it yet. I figured this would be a v1 implementation sort of thing

robgruen:
I figured, I just wanted to call it out so it isn't lost.

ts/packages/dispatcher/dispatcher/src/validation/contextselector/compareLlm.mts
GeorgeNgMsft: The agent descriptions are just an artifact of the previous implementation--it didn't originally have the LLM comparison. When I added it, it was bolted on a minimal setup that only parsed the *.keywords.json files, so it didn't parse the manifest, etc. The authored descriptions were more contrasting than the real agent descriptions, and should be fine, since it doesn't affect the ContextSelector's measurements, it just makes the control look better than the real baseline. 

For the benchmark, I put it in /validation because initially I thought that was where the other benchmark tools/files were located, at a glance. 

I moved it out and am updating the benchmark to pull the real agent descriptions right now (for scalability later on).

PR #2657, opened 2026-07-13
refactor(chat-ui): unify caret placement helper

curtisman:
## Summary

Small refactor in the chat-ui `ChatPanel` input handling that removes a duplicated caret-placement helper and fixes the caret position when recalling commands from history.

## Changes

- Removed the duplicate `moveCursorToEnd()` helper, which was functionally equivalent to the existing `placeCaretAtEnd()`. `acceptCompletion()` now calls `placeCaretAtEnd()`.
- `navigateHistory()` now places the caret at the end of the recalled command text. Previously, after pressing ArrowUp/ArrowDown to recall a prior command, the caret was left at the start of the input, forcing the user to manually move to the end before continuing to edit.

## Rationale

Consolidating the two helpers removes redundancy and keeps a single code path for caret placement. Placing the caret at the end on history recall matches typical shell/terminal behavior and lets users continue typing immediately.


robgruen (approved):

robgruen:
This is a test!

PR #2669, opened 2026-07-14
Harden @package install/update/uninstall and fix npm-provider refcount leak

curtisman:
## Summary

Hardens `@package` install/update/uninstall of installed agents and fixes an npm-provider refcount leak. The user-visible payoff: agent add/remove/update confirmations now render inline in the session, long feed installs stream progress and can be cancelled, and concurrent enables of the same agent no longer leak a duplicate process or desync the refcount.

### npm agent provider: refcount correctness

- Concurrent first-loads of the same agent (e.g. an installed agent enabled in two sessions at once, which share one provider instance) are now coalesced onto a single in-flight load. Previously each load spawned a duplicate agent process and the second insert clobbered the first, leaking a process and desyncing the refcount (two holders, count 1), which later over-unloaded and threw.
- A freshly loaded/created agent now starts at `count: 0`; `loadAppAgent` is the sole place a hold is taken, so the count always reflects the number of holders.

### Installed-agent source: outcomes and messaging

- Added an `unchanged` update outcome for the already-current no-op (a `@package update` on an up-to-date agent is no longer silent); the fan-out cannot express a no-op, so it is reported to the issuing conversation directly.
- Uninstall of a name that is recorded but not `active` in this source now fails loudly instead of silently mutating `agents.json`. In supported deployments (one process per store, one shared source per process) this is an out-of-band inconsistency, not a normal state.
- Extracted the repeated read-modify-write of `agents.json`, the fan-out target set, and the unreferenced-root prune into helpers to remove duplication.

### Consolidated cross-session notification

- A `replaceProvider` swap now emits a single net-effect sibling message (`update` / `remove` / silent on rollback) instead of the raw remove-then-add pair, so an update reads as one "updated" line.
- Agent add/remove/update confirmations are emitted as `Inline` events (not `Info`), so they actually render in the conversation instead of only landing in the shell's notification tray.

### Feed install: progress + cancellation

- Threaded an optional `AbortSignal` through resolve/materialize into the feed source's `npm install` so a long install can be cancelled mid-flight.
- The feed source reports live progress (`Downloading and installing...`, a periodic heartbeat while npm runs, `Finalizing...`) via the `onStatus` callback.
- Install/preview messages now name the source by kind (e.g. `catalog source 'workspace'`) for clearer wording.

## Testing

- New unit tests cover: concurrent first-load coalescing with an honest refcount; the consolidated `notifyReplace` for update / remove / rollback / `notify=false`; `Inline` event emission; the recorded-but-not-active uninstall error; and the uninstall/update command messaging for committed / reverted / unchanged / async-started outcomes.
- Ran the affected suites locally: `provider.spec` (11/11), `appAgentHost.spec` (43/43), and the defaultAgentProvider install-source suites (128/128) all pass.


ts/packages/defaultAgentProvider/src/installSources/config.ts
```diff
@@ -110,6 +110,7 @@ export function deriveMatchKind(m: {
  */
 export interface InstallPreviewMatch {
     readonly source: string;
+    readonly sourceKind?: string; // path / catalog / feed, for the preview message
```
robgruen, line 113: enum?

ts/packages/defaultAgentProvider/src/installSources/feedSource.ts
```diff
@@ -242,7 +246,7 @@ async function defaultNpmInstall(args: NpmInstallArgs): Promise<void> {
             "--userconfig",
             args.userconfig,
         ],
-        { cwd: args.cwd, shell: isWindows },
+        { cwd: args.cwd, shell: isWindows, signal: args.signal },
```
robgruen, line 249: instead of an "isWindows" maybe an os moniker instead?

ts/packages/defaultAgentProvider/src/installSources/config.ts
curtisman: This is abstract... depending on the install factory.  So not fix set of source kind

ts/packages/defaultAgentProvider/src/installSources/feedSource.ts
curtisman: No sure what you mean.   it is based on process.platform already?

robgruen (approved):
Complicated!  Seems like the biggest issue is if the updates stall or fail silently.

curtisman:
> Complicated! Seems like the biggest issue is if the updates stall or fail silently.

yeah.  I want to simplify it in this change, but that would make it even harder for review. it will be a follow up after this PR.

PR #2671, opened 2026-07-14
Add @describe capability to describe agent / actions

GeorgeNgMsft:
### Summary
Adds a new `@describe` command (and matching `system.describe` natural-language schema) that answers **"what can this do?"** — a read-only capability summary, distinct from `@action` which *executes*. It works even for installed-but-disabled agents, so users can discover what an agent offers before enabling it. Intended for new users who may or may not visit our repo and have no familiarity with our agents/actions.

### Usage
- `@describe <agent>` — agent summary + a table of its first 10 actions
- `@describe <agent> --all` / `-a` — full action table, no truncation
- `@describe <agent> <action>` — detailed explanation of one action, including its parameters
- `@describe <action>` — same, when the action name alone is unambiguous across agents
- Natural language: "what can the spotify agent do", "show me all of spotify's actions", "what does the play action do"

Agent Describe Command
image

Agent Describe NL
image

Action Description
image


curtisman:
there is also the `@help` command as well. that serve similar function for commands.... should all of them under one command?

ts/packages/dispatcher/dispatcher/src/context/system/describe/agentSchemaInfo.ts
```diff
@@ -0,0 +1,99 @@
...
+import { getAppAgentName } from "../../../translation/agentTranslators.js";
+
+/** Extract action names + descriptions from a parsed action schema. */
+export function extractActions(
+    actionSchemas: Map<string, ActionSchemaTypeDefinition>,
+): ActionInfo[] {
+    const actions: ActionInfo[] = [];
+    for (const [name, actionDef] of actionSchemas) {
+        const description =
+            (actionDef.comments?.[0] ?? "").trim() ||
```
curtisman: may be a `getActionDescription` helper that can share with `unknownSwitcher.ts`?

ts/packages/dispatcher/dispatcher/src/context/system/describe/agentSchemaInfo.ts
```diff
@@ -0,0 +1,99 @@
...
+            name
+                .replace(/([A-Z])/g, " $1")
+                .replace(/^./, (c) => c.toUpperCase())
+                .trim();
```
curtisman: is the fall back useful?

ts/packages/dispatcher/dispatcher/src/context/system/action/describeActionHandler.ts
```diff
@@ -0,0 +1,59 @@
...
+export async function executeDescribeAction(
+    action: AppAction,
+    context: ActionContext<CommandHandlerContext>,
+) {
+    const nlAction = action as unknown as DescribeAction;
+    switch (nlAction.actionName) {
+        case "describeAgent": {
+            const { agentName, all } = nlAction.parameters;
+            await processCommandNoLock(
+                `@describe ${quoteArg(agentName)}${all ? " --all" : ""}`,
```
curtisman: may be "implicitQuote" arg might be of use here?

curtisman (approved):

GeorgeNgMsft:
> there is also the `@help` command as well. that serve similar function for commands.... should all of them under one command?

I've been thinking about it and I do agree that instinctively @help feels like it should be merged. They function differently under the hood and operate on different sources (command tree vs action schema). I might do a follow up PR that adds a similar NL rule to allow @help to function the same way and let @describe also describe the agent commands. But I'm not entirely convinced yet that we should merge their scopes

PR #2678, opened 2026-07-15
Improve test reliability and add test mode options

robgruen:
This pull request contains several improvements to the test infrastructure and reliability for the shell and agent provider packages. The changes focus on making test runs more robust, preventing race conditions, and ensuring that tests do not fail due to slow shutdowns or interference from other running instances.

**Test infrastructure and reliability improvements:**

* Improved the `@session clear` test to wait for the session reinitialization to fully complete and verify the success message before shutting down, preventing shutdown race conditions that could cause test timeouts. [[1]](diffhunk://#diff-e96c3d0f3434e1243fc6ddf8cc0bf5cd95a5d21db7bf4c6652bcfed0af101391R158-R179) [[2]](diffhunk://#diff-e96c3d0f3434e1243fc6ddf8cc0bf5cd95a5d21db7bf4c6652bcfed0af101391R7)
* Enhanced the `closeInstance` helper to force-kill the process and log a warning if graceful shutdown exceeds the timeout, instead of failing the test, making tests more resilient to slow or hung shutdowns.

### DIFF
--- a/ts/packages/shell/test/testHelper.ts
+++ b/ts/packages/shell/test/testHelper.ts
@@ -212,9 +212,7 @@ export async function exitAndAwaitCleanShutdown(
     const instanceName = process.env["INSTANCE_NAME"]!;
     const existing = runningApplications.get(instanceName);
     if (existing === undefined) {
-        throw new Error(
-            `No running instance '${instanceName}' to shut down.`,
-        );
+        throw new Error(`No running instance '${instanceName}' to shut down.`);
     }
     const proc = existing.process();
     const start = Date.now();
--- a/ts/packages/shell/test/testHelper.ts
+++ b/ts/packages/shell/test/testHelper.ts
@@ -82,7 +82,9 @@ async function closeInstance(instanceName: string, force: boolean = false) {
 /**
  * Starts the electron app and returns the main page after the greeting agent message has been posted.
  */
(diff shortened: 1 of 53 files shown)


ts/packages/shell/test/testHelper.ts
```diff
@@ -52,13 +52,23 @@ async function closeInstance(instanceName: string, force: boolean = false) {
...
             await waitForPromiseWithTimeout(existing.close(), 10000);
         } catch (e: any) {
-            const errMsg = `Failed to close instance ${instanceName}: ${e.message}.\nKilling instance ${instanceName} PID: ${existing.process().pid}`;
-
-            existing.process().kill();
-            if (force) {
-                console.log(errMsg);
-            } else {
-                throw new Error(errMsg);
+            // A graceful close() can legitimately exceed the timeout when the
```
curtisman, line 55: I am a little bit worried about this hiding problems?   Do we have a test to make sure a clean "exit" is possible?

curtisman (approved):

ts/packages/shell/test/testHelper.ts
robgruen: I will add one.  This is a fallback.  The other fix in this PR is to resolve us hitting this.  Basically, we were shutting down during a new session creation which ultimately led to getting here.  Now that that is fixed we won't need this.

PR #2686, opened 2026-07-16
Make ContextSelector grammar cache compatible

GeorgeNgMsft:
Make  contextSelector  work with the construction cache (registry-first Tier 1.5)

**Summary**

Today, when a phrase could belong to two agents but the construction cache has already learned it for just one, the router sees a single confident answer and stops. The topic-based resolver ( contextSelector ) never gets to weigh in, so it can't fix the routing even when recent conversation makes the right agent obvious. This PR lets it step in for those "cache-hidden" collisions. (Before, you had to turn the cache off with  @const builtin off  just to see the problem.)

**Why**

The cache returns one match for an ambiguous phrase, but  contextSelector  needs at least two options to choose between. So the cached answer always wins by default. The fix: use the neighborhood registry to notice the hidden ambiguity, then let  contextSelector  pick the right agent from it.

**What changed**

image

• New step (Tier 1.5) in  resolveGrammarRegistryFirst  ( matchCollision.ts ): when the registry marks a single cache match as ambiguous, it pulls in the related "sibling" agents and lets  contextSelector  choose among them. It runs after an explicit user choice (Tiers 0 and 1), so it never overrides one, and before asking the user to clarify (Tier 2) — so when intent is clear, it just routes automatically instead of prompting.
• Scorer cleanup ( matchContextSelector.ts ): a shared  scoreAndDecide  core, plus a new  resolveContextSelectorMembers  entry point that can score the registry siblings (which don't carry a  MatchResult ) next to the cache matches.
• Routing: if a cache match wins, it routes with no LLM call. If a sibling wins, it saves a one-turn hint ( pendingTopicalRoute , a new field) that  pickInitialSchema  reads to lock in the agent and show the  ↪ routed to … — recent topic  note — only once the route is final, and it works even with embeddings off. This replaces an older hint that lived too long, so the note can't be wrong or carry over into the next turn.
• Safety limits: siblings are limited to agents active this turn (so it never routes to a disabled one), and it scores every real match plus the registry siblings together (so a genuine two-way match never loses a valid option).

**Gating**

Off by default and does nothing unless you opt in. It only runs with  preference.registryFirst: on  and a loaded  preference.registryPath  (both off/empty by default) and  contextSelector.detect  (on by default). Default config is unchanged.

**Accepted trade-off**

With  registryFirst  on and a real two-way match, scoring includes the registry siblings too — so a sibling the cache didn't match, but that fits the recent topic well, can win. This can change the outcome compared to scoring only the cache matches. That's on purpose: the cache isn't the final word, and a strong topic signal should be able to correct it. It stays safe because the evidence bar leans toward asking the user — more contenders means it's more likely to clarify, not less.


robgruen:
We are going to have trouble assessing the quality/usefulness of this until we actually have real users with real behaviors and preferences.

GeorgeNgMsft:
> We are going to have trouble assessing the quality/usefulness of this until we actually have real users with real behaviors and preferences.

I agree, I implemented this though so that it could be compatible with grammar caching, which would quickly skip it otherwise

ts/packages/dispatcher/dispatcher/src/translation/matchCollision.ts
```diff
@@ -99,21 +101,36 @@ export function isCollision(
     );
 }
 
+// Find the validated cache match whose primary (schema, action) equals `member`,
+// or undefined when `member` is a registry sibling the grammar didn't produce.
+function findValidatedByMember(
```
robgruen, line 106: Just putting this comment here since I just had this thought.

Routing/action selection is getting overly complex, it was very simple in the beginning (cache hit or translate).  It has evolved quite significantly.

It would be great if we could experiment with gathering all of these signals (they are quick to execute: grammar, cache, context, collisions, preferences, etc.). And then have another mechanism that decides the winner.  Right now, there's a sequence of checks, comparisons, etc. that is increasing with complexity with every check-in.  

I would think it might be possible to collect all of these signals and *then* decide amongst them.

robgruen: Basically what I'm thinking is "vote by committee".

robgruen: The more of this code I look at the more I'm convinced that this is just an abstraction of "translation".  Instead of the candidate actions being the set of all available agent actions we are now using these different preference mechanisms to strip the action space into a more reasonable representation of what the users' intent is.

So maybe a better architecture is that these action selection mechanisms can be registered with the translator (LLM is just one method): grammar, preference, context, collision, etc. (i.e. turn them on or off).  And then once you've collected all the different actions from these providers you pick the winner.  We're essentially ranking them.  

It's only a matter of time before we dream of another way to disambiguate and that then becomes another fork in the code.

robgruen (approved):

ts/packages/dispatcher/dispatcher/src/translation/matchCollision.ts
GeorgeNgMsft: This is a great idea, I'll mark it down as a follow up

PR #2688, opened 2026-07-17
Streamline electron package

hillary-mutisya:
- Switch default behavior for shell launch to connect to a running agent-server instance. If none is running, the shell will launch one, matching the behavior of CLI.
- Add `--standalone` flag which alllows launching shell with in-proc hosted agent-server, for dev environments that want previous behavior.
- Update electron packaging to only include platform-specific packages e.g. only x64 copilot sdk packages when building x64.
- Mark the onnxruntime components used for local embedding as optional. Remove them from shell package since emebeddings would run in agent-server.
- Together, the last 2 changes cut 40% of the shell installer size.


robgruen (approved):

hillary-mutisya:
@copilot resolve the merge conflicts in this pull request

curtisman:
@hillary-mutisya Does the `@shell` agent works in server mode?   I don't think support for "per client" agent injection is implemented yet.   Might need a follow up work item.

hillary-mutisya:
@curtisman , good call. I'm tracking the agent registration as a separate change (https://github.com/microsoft/TypeAgent/pull/2697). I'll make sure that PR goes in before this one.

shell:
I do work in agent mode. Reluctantly

PR #2689, opened 2026-07-17
TypeAgent Studio: Trace Viewer - an A/B resolution-pipeline debugger for utterance regressions

TalZaccai:
## What & why
When the Impact Report flags an utterance that resolved differently between two
versions (A = base, B = compare), this adds a **Trace Viewer** that answers
*where* resolution diverged and *what change* caused it.

For one utterance it renders resolution top-to-bottom as a pipeline —
construction cache → grammar match → wildcard validation → result — for both
versions. Agreeing stages collapse to one line; the first diverging stage expands
into a side-by-side A/B card, names the responsible grammar/schema file, and
offers an inline A↔B file diff plus jump-to-source. Opened in one click from an
Impact Report row (right-click, detail-pane button, or arrow-key nav).

## Design decisions worth a look
- **Parity is decided by comparing the two sides' final actions, not per-layer
  outcomes.** Both sides can "hit" yet produce different actions, and a live
  cache hit vs a git-ref grammar match are different branches, not the same step.
- **Cause attribution walks layers in order** and blames the first that explains
  the difference (cache short-circuit → different grammar rule → wildcard
  rejection → action payload). Grammar blame uses the compiled-grammar **hash**
  (resolver-independent); when the captured parse can't be trusted, attribution
  falls back to the action level and the conclusion is marked **low-confidence**.
- **Tracing is side-effect-free on the hot path** — `resolveWithTrace()` shares
  `resolveSide()` with the normal `resolve()`, so uninstrumented replay pays
  nothing.
- **Heavy traces stay out of the webview payload** — the report posts only a
  lightweight `tracedUtteranceIds` list (MB → KB); full traces are persisted
  separately and loaded on open. Per-run capture is changed-first and capped.

## Where the changes live
- **`typeagent-core`** — produces/captures/persists/recomputes traces: trace
  DTOs, resolver instrumentation (`grammarResolver`,
  `constructionCacheResolver.matchEntry`), per-row capture + single-utterance
  recompute (`studioRuntimeCore`).
- **`typeagent-studio`** — the viewer: A/B pipeline client + host (owns all
  file/diff opening), the pure divergence view model, source resolver,
  workspace-state trace store (retention/eviction), and the Impact Report entry
  points.
- **`grammarTools`** — `gt-trace-timeline` gains a display-only, CSP-safe mode so
  a persisted match can render inside a webview.
- **`studio-service`** — one RPC handler bridging the recompute call.

## Testing
Divergence view model, source resolver, trace store, protocol, engine
capture/recompute, and the timeline display-only path are all covered; full
`typeagent-studio` and `typeagent-core` suites green.


ts/packages/typeagent-core/src/replay/grammarResolver.ts
```diff
@@ -507,98 +578,366 @@ export function createGrammarReplayResolver(
...
+    const actionName =
+        typeof rec?.actionName === "string" ? rec.actionName : undefined;
+    const sourceFilePath = target.schema?.sourceFilePath;
+    const schema =
+        sourceFilePath !== undefined || actionName !== undefined
+            ? {
+                  ...(sourceFilePath !== undefined ? { sourceFilePath } : {}),
+                  ...(actionName !== undefined ? { actionName } : {}),
+              }
+            : undefined;
```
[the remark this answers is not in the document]
TalZaccai: Good catch. Fixed in a54b84314: actionNode now attaches schema.sourceFilePath only when the side actually produced an action, so a miss no longer offers a schema jump or A/B compare. Added a regression test covering both the miss (no schema) and hit (schema present) cases.

ts/packages/typeagent-studio/src/traceViewerView.ts
```diff
@@ -0,0 +1,485 @@
...
+        return undefined;
+    }
+    if (gitRoot.length === 0) {
+        return undefined;
+    }
+    const relPath = path.relative(gitRoot, real).split(path.sep).join("/");
+    if (relPath.startsWith("..") || path.isAbsolute(relPath)) {
+        return undefined;
+    }
+    return { gitRoot, relPath };
```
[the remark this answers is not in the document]
TalZaccai: Fixed in a54b84314: the guard now rejects only a real ".." path segment (relPath equals ".." or starts with "../") instead of any path beginning with two dots, so an in-repo file named like "..foo" is allowed through.

ts/packages/typeagent-studio/src/webviewKit/client/impactReport.ts
```diff
@@ -115,13 +115,20 @@ let currentTotal = 0;
 // Raw deltas of the current result keyed by utterance id, so a row drill-in can
 // build the action A/B diff without the host re-sending the payload.
 let currentRawById = new Map<string, ActionDelta>();
+// Utterance ids that captured a resolution trace this run (only red rows do), so
+// the drill-in offers "Open trace" exactly for the rows the viewer can show.
+let tracedIds: Set<string> = new Set();
```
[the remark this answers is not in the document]
TalZaccai: Fixed in a54b84314: reworded the comment to describe the current capture-all behavior (changed rows first, then equal rows up to the cap) rather than the stale red-rows-only wording.

ts/packages/typeagent-studio/src/webviewKit/client/impactReport.ts
```diff
@@ -1239,46 +1387,54 @@ function closeDetail(): void {
     detailEl.textContent = "";
 }
 
-/** Paint the detail pane: a header (utterance + close) and the unified A/B diff
- *  of the two resolved actions. */
+/** Paint the detail pane: a header (utterance + optional trace drill-in +
+ *  close) and the unified A/B diff of the two resolved actions. */
+/** Paint the detail pane: a header (utterance + close) and, below it, the two
+ *  resolved actions as a unified A/B diff — which for an equal row is the
+ *  resolved action JSON shown as all-context lines. */
```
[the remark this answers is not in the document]
TalZaccai: Fixed in a54b84314: removed the duplicated JSDoc block so renderDetail now carries a single description of its current behavior.

ts/packages/grammarTools/ui/src/gt-trace-timeline.ts
```diff
@@ -316,6 +345,36 @@ export class GtTraceTimeline extends LitElement {
         }
     }
 
+    /** True when the component shows a supplied stored trace instead of running
```
robgruen, line 359: Is there a way we can stop copilot from embedding HTML and instead use a template/replacement pattern?

ts/packages/typeagent-core/src/replay/grammarResolver.ts
```diff
@@ -55,7 +55,7 @@ import {
     type AgentRootsInput,
 } from "../health/index.js";
 import type { CorpusEntry } from "../corpus/types.js";
-import type { VersionSpec } from "./types.js";
+import type { VersionSpec, ReplayCacheState } from "./types.js";
```
robgruen, line 58: As this evolves consider if some/part/all of this can be folded into the grammar toolset as a first class citizen.

ts/packages/typeagent-studio/src/webviewKit/client/traceViewer.ts
```diff
@@ -0,0 +1,888 @@
...
+                sourceNote = msg.message;
+            }
+            render();
+            break;
+    }
+});
+
+vscode.postMessage({ type: "ready" });
+
+// --- Rendering ------------------------------------------------------------
```
robgruen: this seems silly..if there's a logical break in the logic of the file shouldn't we just split the file and name them accordingly?

ts/packages/typeagent-studio/src/webviewKit/client/traceViewer.ts
```diff
@@ -0,0 +1,888 @@
...
+        tip.push(`Namespace: ${cache.namespace}`);
+    }
+    if (cache.parts !== undefined && cache.parts.length > 0) {
+        tip.push(`Pattern: ${cache.parts.join(" ")}`);
+    }
+    id.title = tip.length > 0 ? tip.join("\n") : "Matched construction id.";
+    return id;
+}
+
+// --- State screens & bits -------------------------------------------------
```
robgruen: Yeah, let's refactor this into it's logical parts.

ts/packages/typeagent-studio/src/traceStore.ts
```diff
@@ -0,0 +1,235 @@
...
+// Per-red-row resolution traces + the run descriptor are persisted here so the
+// Trace Viewer can open the exact side-by-side resolution that produced a row
+// long after the run streamed, and can recompute a fresh trace from the same
+// pinned inputs. Retention mirrors the Impact Report: one run is kept per agent
+// (a new run for the agent evicts the previous one), and a small global cap
+// bounds how many agents' runs are retained at once.
+
+/** Keep the last run for at most this many agents so the persisted payload stays
+ *  bounded even if the user cycles through many agents. */
+const MAX_RETAINED_RUNS = 8;
```
robgruen: configuration setting?

robgruen (approved):

ts/packages/typeagent-studio/src/traceStore.ts
TalZaccai: Done in b6bc9c8d8: `MAX_RETAINED_RUNS` is now the `typeagentStudio.traceViewer.maxRetainedRuns` setting (integer, minimum 1, default 8). `traceStore.saveTraceRun` takes the cap as a parameter and `impactReportView` reads it from configuration.

ts/packages/typeagent-studio/src/webviewKit/client/traceViewer.ts
TalZaccai: Done in b6bc9c8d8: split `traceViewer.ts` into logically-named sibling modules — `traceViewerDom`, `traceViewerBridge`, `traceHeader`, `traceStateScreens`, `traceStageParts`, `traceCause`, `tracePipeline` — leaving `traceViewer.ts` as a slim entry (state + message loop + render).

ts/packages/typeagent-studio/src/webviewKit/client/traceViewer.ts
TalZaccai: Done in b6bc9c8d8: this block moved into its logical modules — the stage/cause/pipeline builders now live in `traceStageParts`, `traceCause` and `tracePipeline`, and `traceViewer.ts` keeps only the entry wiring.

ts/packages/grammarTools/ui/src/gt-trace-timeline.ts
TalZaccai: Done in b6bc9c8d8: the monolithic `render()` is decomposed into template-returning helper methods (`_renderControls`, `_renderSummaryBar`, `_renderFilterBar`, `_renderTraceTable`, `_renderTraceRow`), so `render()` now composes small named `lit` templates instead of one large inline block.

ts/packages/typeagent-core/src/replay/grammarResolver.ts
TalZaccai: Agreed — this replay resolver overlaps a lot with the grammar toolset's own matching and debug surface. Leaving it as-is for this PR, but I'll track folding the shared pieces (grammar compile + trace capture) into the toolset as a first-class citizen as a follow-up.

PR #2697, opened 2026-07-18
Add client-hosted agent registration

hillary-mutisya:
Enable clients to register local app agents with a conversation dispatcher. Adds registerClientAgent/unregisterClientAgent RPCs and protocol types, client-side agent-rpc server lifecycle management, and server-side agent-rpc client handling to install/remove dynamic agents on a conversation. Updates conversation manager and shared dispatcher to add/remove client-hosted agents, threads cleanup on reconnect/close, and wires shell to register its agent when connecting (including createShellAgent helper). Also adds @typeagent/agent-sdk dependency where needed and updates test stubs and service-worker adapter to no-op these calls.


GeorgeNgMsft (approved):

ts/packages/agentServer/server/src/conversationManager.ts
```diff
@@ -846,6 +860,39 @@ export async function createConversationManager(
...
+            conversationId: string,
+            name: string,
+            manifest: AppAgentManifest,
+            appAgent: AppAgent,
+        ): Promise<void> {
+            const record = conversations.get(conversationId);
+            if (record === undefined) {
+                throw new Error(`Conversation not found: ${conversationId}`);
+            }
+            const sharedDispatcher = await ensureDispatcher(record);
```
curtisman, line 873: @hillary-mutisya does this mean that all the client connected to the same conversation will have "shell" action available?   Is that the right behavior?

ts/packages/agentServer/server/src/conversationManager.ts
```diff
@@ -846,6 +860,39 @@ export async function createConversationManager(
...
+            conversationId: string,
+            name: string,
+        ): Promise<void> {
+            const record = conversations.get(conversationId);
+            // If the conversation or its dispatcher is already gone, there is
+            // nothing to remove.
+            if (record?.sharedDispatcher === undefined) {
+                return;
+            }
+            await record.sharedDispatcher.removeDynamicAgent(name);
```
curtisman, line 890: what happens if multiple shell client connects?

PR #2705, opened 2026-07-20
Prevent duplicate values in MultiMap.addUnique

datduyng:
A case-insensitive alias index duplicates each new unique value after the first:

WRONG
```console
> const aliases = new MultiMap(); const sameAlias = (a, b) => a.toLowerCase() === b.toLowerCase()
> aliases.addUnique("users", "Alice", sameAlias).addUnique("users", "Bob", sameAlias)
MultiMap(1) { 'users' => [ 'Alice', 'Bob', 'Bob' ] }
> aliases.get("users")
[ 'Alice', 'Bob', 'Bob' ]
```


ts/packages/typeagent/test/lib.array.spec.ts
```diff
@@ -3,6 +3,28 @@
...
+describe("collections.MultiMap", () => {
+    test("addUnique with comparer inserts each value once", () => {
+        const map = new collections.MultiMap<string, { id: number }>();
+        const sameId = (x: { id: number }, y: { id: number }) => x.id === y.id;
+
+        map.addUnique("key", { id: 1 }, sameId);
+        map.addUnique("key", { id: 2 }, sameId);
+        map.addUnique("key", { id: 2 }, sameId);
+
+        expect(map.get("key")).toEqual([{ id: 1 }, { id: 2 }]);
```
robgruen, line 15: Maybe after this do an insert with the duplicate key to make sure the id is actually updated.

robgruen (approved):

robgruen:
Nice find!

PR #2712, opened 2026-07-21
Fix short flags being consumed as string values

datduyng:
A string-valued short flag could incorrectly consume the next short flag as its value. Here, `-m` expects a match string while `-v` is a separate boolean flag.

```console
> @const list -m -v

Input:
  -m -> expects a string value
  -v -> verbose boolean flag

Before:
  parsed match -> "-v"
  parsed verbose -> false
  output -> no result

After:
  -v is recognized as another flag, leaving -m without a value
  output -> Missing value for flag '-m'
```

The short-flag check used an impossible condition (`character < "0" && character > "9"`). Changing it to `||` rejects unquoted flag tokens as string values while still allowing negative numeric values such as `-1` and quoted values such as `'-v'`.

https://github.com/user-attachments/assets/23fc609d-8c9c-443b-9ced-f605387b1662

### DIFF
--- a/ts/packages/dispatcher/dispatcher/src/command/parameters.ts
+++ b/ts/packages/dispatcher/dispatcher/src/command/parameters.ts
@@ -85,8 +85,8 @@ function invalidValueToken(valueToken: string) {
     if (
         valueToken.length === 2 &&
         valueToken[0] === "-" &&
-        valueToken.charCodeAt(1) < 48 /* "0" */ &&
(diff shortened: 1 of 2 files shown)


datduyng:
https://github.com/user-attachments/assets/23fc609d-8c9c-443b-9ced-f605387b1662

GeorgeNgMsft (approved):

PR #2713, opened 2026-07-21
[VS Code Shell] Improve Reasoning Action UI

GeorgeNgMsft:
**Improve  dispatcher.reasoningAction  UI + expandable messages**

Improvements to how reasoning responses render in the TypeAgent chat, plus a way to pop a message out into its own window. All UI logic lives in the shared  chat-ui  package, so both the VS Code shell and the Electron shell benefit.

What changed

**Reasoning UI**

• Auto-collapse completed "Thinking" blocks. While a reasoning step is active it stays expanded; once superseded or the request completes, it details collapses so the finished trail minimizes. Collapse is per-block (WeakSet-guarded), so a block you manually reopen isn't slammed shut again.
• Purple "working" rail for reasoning. The status rail reads light purple (instead of the generic blue) while a  dispatcher.reasoningAction.*  request is in flight.
• List/heading spacing fixed. A host CSS reset had stripped markdown list indent and heading margins; restored them, with bullet/number markers aligned to the paragraph text.

image

image

**Expand message**

• New maximize button in each message's action row (between copy and 👍).
• In the VS Code shell it opens the message in a separate editor panel you can move / split / snap to a new window. The panel HTML is static; content is delivered over  postMessage  and re-sanitized with DOMPurify in a small dedicated webview ( expand.js ) — no extension-side HTML injection. Strict CSP, links routed externally, paginated tables fully expanded for reading.
• Other hosts (Electron / Chrome) fall back to an in-page overlay that moves the live nodes (keeping links/tables interactive). Iframe-backed messages always use the overlay.
• Fixed the panel tab icon being invisible on dark themes (theme-aware  { light, dark }  robot icon).

image


ts/packages/chat-ui/src/feedbackWidget.ts
```diff
@@ -311,6 +333,67 @@ function makeIconButton(
...
+// The single expand overlay currently open (its disposer), so a second click
+// replaces rather than stacks overlays.
+let activeExpand: (() => void) | undefined;
+
+// Open a message's rendered content in a full-window overlay for easier reading
+// of long responses. The live content nodes are moved into the overlay (not
+// cloned) so links, tables and other interactivity keep working, then moved
+// back when it closes. Dismissed by the close button, Escape, or a backdrop
+// click.
+function openMessageExpand(sourceEl: HTMLElement) {
```
GeorgeNgMsft: I think I'll rename this, as it was previous an expansion capability but now it's really openMessageInNewWindow

GeorgeNgMsft: And I'll cleanup some of the naming/other files around it. "Expansion" has been conflated with "new window" in a number of places

hillary-mutisya (approved):

ts/packages/chat-ui/src/feedbackWidget.ts
GeorgeNgMsft: Actually, I'm going to do a bit more of a refactor since some of this feature has been shared between Chat UI when it really shouldn't at the moment (made sense previously). Will slim it down to just what's necessary

GeorgeNgMsft: Refactored and cleaned up

PR #2716, opened 2026-07-21
Shell / vscode-shell bugfixes

TalZaccai:
# Shell / vscode-shell bugfixes

- **Action templates**: `string-union` params no longer crash template build; `any`/`true`/`false`/`undefined` skipped gracefully.
- **`@notify` in vscode-shell**: fixed broken subcommands by moving the notification buffer into shared `chat-ui`; fixed the `@notify show` bullet overlap.
- **`@open` in vscode-shell**: now opens the folder in the OS file manager (was silently dropped).
- **Status notices**: minimizing a notice now collapses it to a bell next to the connection indicator (click to reopen); the notice card is theme-aware (dark card + colored accent in dark mode, pastel fallback elsewhere).
  - `@notify status` message: image
  - Collapsed view: image


ts/packages/chat-ui/src/chatPanel.ts
```diff
@@ -2882,6 +3027,64 @@ export class ChatPanel {
...
+     * toast/inline/info/warning/error events so the notification center has
+     * content to summarize. Purely UI state — no host dependency.
+     */
+    public recordNotification(
+        event: string,
+        source: string,
+        data: unknown,
+    ): void {
+        this.notifications.push({ event, source, data, read: false });
+    }
```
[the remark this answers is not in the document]
TalZaccai: Good catch — capped the buffer at 500 entries in recordNotification(); once it's full the oldest entries are dropped so it can't grow without bound. Fixed in 9235644dd.

ts/packages/chat-ui/src/chatPanel.ts
```diff
@@ -2882,6 +3027,64 @@ export class ChatPanel {
...
+                const unread = this.notifications.filter((n) => !n.read).length;
+                this.showInline(
+                    {
+                        type: "html",
+                        content: `There are <b>${unread}</b> unread and <b>${this.notifications.length}</b> total notifications.`,
+                    },
+                    "shell",
+                );
+                break;
+            }
```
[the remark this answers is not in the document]
TalZaccai: Thanks. The html content does get sanitized by DOMPurify at the setContent() sink, so this wasn't a script/attribute-injection hole — but a literal '<' in the event/data could still mangle the rendered text, so I now HTML-escape both event and data before interpolating. I also dropped the hard-coded 'shell' source label so the row reads correctly in every host (it was misleading in VS Code). Fixed in 9235644dd.

ts/packages/vscode-shell/src/bridge/clientIO.ts
```diff
@@ -310,6 +310,20 @@ export function createBridgeClientIO(ctx: BridgeClientIOContext): ClientIO {
                         );
                     },
                 );
+            } else if (action === "open-folder") {
```
robgruen, line 313: are we missing any other functionality that's in the shell but not in the vscode-shell?

ts/packages/vscode-shell/src/webview/main.ts
```diff
@@ -251,6 +251,22 @@ chatPanel.onDemoAction = (action: "continue" | "cancel") => {
     vscode.postMessage({ type: "demoCommand", action });
 };
 
+// Render the collapsed status-notice affordance as a bell next to the
```
robgruen, line 254: One of my mock designs had this to the left side of the header bar.  Would be nice to include some screen shots as part of the PR and or retain mockups in the repo that we can reference later.  Hard to envision from text alone.

robgruen (approved):

ts/packages/vscode-shell/src/bridge/clientIO.ts
TalZaccai: Most of it you were already addressing from what I could find

ts/packages/vscode-shell/src/webview/main.ts
TalZaccai: Added screenshots :)

PR #2728, opened 2026-07-23
Show warnings from successful grammar loads

datduyng:
Displays existing validation warnings after a successful grammar load.

### DIFF
--- a/ts/packages/grammarTools/cli/src/cli.ts
+++ b/ts/packages/grammarTools/cli/src/cli.ts
@@ -56,6 +56,7 @@ async function main(): Promise<void> {
                             {
                                 ok: true,
                                 rules: g.identifiers.ruleIds.length,
+                                diagnostics: result.diagnostics,
                             },
                             null,
                             2,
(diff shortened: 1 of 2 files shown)


ts/packages/grammarTools/cli/src/cli.ts
```diff
@@ -65,6 +65,9 @@ async function main(): Promise<void> {
                     console.log(
                         `Loaded: ${g.identifiers.ruleIds.length} rules`,
                     );
+                    for (const diagnostic of result.diagnostics ?? []) {
+                        console.warn(diagnostic.message);
```
hillary-mutisya, line 69: You probably want to use the debug package for all the diagnostic logging. Debug allows us to switch specific namespaces on/off when we need to. Raw console logs are always emitted though, which can be noisy

hillary-mutisya (approved):

ts/packages/grammarTools/cli/src/cli.ts
datduyng: Discussed offline - Hillary is okay with this as diagnostic in this case is warning only

PR #2731, opened 2026-07-23
Add Android chat sample

jebrans:
## Summary
Add a reusable Android Jetpack Compose sample for TypeAgent that connects to a local agent-server over WebSocket and renders a minimal chat UI with streamed assistant responses.

## Example
**Before**
No Android sample was available for app developers to copy and run locally.

**After**
A new `android/samples/mobile-2` sample is available with:
- message history
- connection status
- text input and Send button
- streamed assistant replies in one bubble
- emulato
Screenshot 2026-07-22 115936
r-friendly local server connection

## Why this is needed
This gives Android developers a working reference for integrating TypeAgent into a mobile app instead of starting from scratch. It also documents the local emulator setup (`ws://10.0.2.2:8999`) and the TypeAgent WebSocket flow in a concrete, runnable sample.


android/samples/mobile-2/app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <domain-config cleartextTrafficPermitted="true">
+        <domain>10.0.2.2</domain>
+        <domain>localhost</domain>
```
hillary-mutisya, line 5: This is OK for now, but we should ensure we can authenticate and connect to agent-service running on a separate machine. That is required before we can run this code on an actual android phone or tablet (where agent-server will not be on localhost). Review TypeAgent/ts/examples/remoteClient at main · microsoft/TypeAgent for a quick example on setting up and connecting to the dev tunnel

hillary-mutisya (approved):

android/samples/mobile-2/app/src/main/res/xml/network_security_config.xml
jebrans: Understood, yes this was an implementation to connect locally. I will follow up with remoteClient example

PR #2744, opened 2026-07-26
Enhance agent discovery with new registries and improved descriptions

hillary-mutisya:
This pull request updates the descriptions for several agent `package.json` and manifest files to provide clearer and more informative summaries of each agent's purpose and functionality. These improvements enhance readability and make it easier for developers and users to understand what each agent does at a glance.

### DIFF
--- a/ts/packages/agents/test/package.json
+++ b/ts/packages/agents/test/package.json
@@ -2,7 +2,7 @@
   "name": "test-agent",
   "version": "0.0.1",
   "private": true,
-  "description": "Dispatch agent for testing",
+  "description": "Agent to do addition",
   "keywords": [
     "typeagent-agent"
   ],
--- a/ts/packages/agents/androidMobile/package.json
+++ b/ts/packages/agents/androidMobile/package.json
@@ -2,7 +2,7 @@
   "name": "android-mobile-agent",
   "version": "0.0.1",
   "private": true,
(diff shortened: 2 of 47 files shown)


ts/packages/agents/test/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "test-agent",
   "version": "0.0.1",
   "private": true,
-  "description": "Dispatch agent for testing",
+  "description": "Agent to do addition",
```
robgruen, line 5: This one seems incorrect?

robgruen (approved):

ts/packages/agents/test/package.json
jebrans, line 5: Is this correct? Do we need this agent since we only have a math operation

hillary-mutisya: Yeah, I picked the agent description from its manifest (ts/packages/agents/test/src/manifest.json). It's a test agent - I've not traced which test scenarios use it, so I will keep it for now.

jebrans (approved):

ts/packages/agents/test/package.json
hillary-mutisya: in this change, I am trying to keep the descripion in package.json aligned with that in the agent manifes. In this case, the description was read from ts/packages/agents/test/src/manifest.json.

PR #2754, opened 2026-07-29
TypeAgent Studio: New agent wizard

TalZaccai:
## Summary
Adds the **New Agent wizard** to TypeAgent Studio: a VS Code webview that walks a user from an API
source (doc URL / OpenAPI spec / CLI command) to a working, packaged agent through 7 phases —
Discovery → PhraseGen → SchemaGen → GrammarGen → Scaffolder → Testing → Packaging.
Command: `typeagent-studio.newAgent`.

## What this PR does
- **Wizard UI.** A new multi-phase webview in the extension that drives the whole onboarding flow and
  streams per-phase progress, ending with a "Try it" step.
- **Onboarding channelization.** Onboarding now runs in the studio-service and communicates over RPC
  channels; the extension's in-process dispatcher is removed. A service-side `phaseRunner` orchestrates
  the 7 phases.
- **Real build before Testing/Packaging.** The scaffolded agent is now compiled during the Scaffolder
  phase, so Testing and Packaging operate on a built agent (fixes the `ERR_MODULE_NOT_FOUND` gap). The
  build runs isolated (`--ignore-workspace`) so onboarding never triggers a workspace-wide install.
- **"Try it" proof.** After scaffolding, the generated agent is path-loaded and an utterance is
  *translated* (not executed) to prove the wizard produced a routable agent.
- **Reliability fixes.** GrammarGen gained a repair loop + sanitizers so LLM grammars compile
  deterministically, and the CLI handler codegen template was fixed to type-check.
- **Demo target.** A small .NET `randomStub` CLI (with a one-command deploy script) serves as a
  source-controlled Discovery target for the walkthrough.

## Testing
- New specs: `onboardingPhaseRunner`, `utteranceProver`, `generatedAgentTranslator`, `wizardProtocol`,
  `wizardViewModel`, webview bundle.
- End-to-end in the dev host against `randomStub`: all 7 phases green (Testing 75/75, Packaging validated).

## File breakdown
- **Wizard UI** (`typeagent-studio`): `wizardView.ts`, `webviewKit/{wizardProtocol,wizardViewModel}.ts`,
  `webviewKit/client/wizard.ts`, `media/wizard.css`, `esbuild.mjs`; command wiring in `commands.ts` /
  `extension.ts`; **deletes** `studioRuntime.ts`; talks to service via `serviceRuntimeFacade.ts` /
  `studioServiceClient.ts`.
- **Channelization** (`typeagent-core` + `studio-service`): onboarding channels in `studioProtocol.ts` /
  `studioRuntimeCore.ts`; `onboarding/{phaseRunner,dispatcherGateway,utteranceProver}.ts`; wired via
  `runtime.ts` / `studioRpcHandlers.ts`.
- **Agent loading** (`defaultAgentProvider`): `onboardingDispatcher.ts`, `generatedAgentTranslator.ts`,
  `serviceKeys.ts`.
- **Pipeline** (`agents/onboarding`): `lib/agentBuild.ts`, `scaffolderHandler.ts` (build-in-scaffolder +
  isolated build), `grammarGenHandler.ts` (repair loop), `cliHandler.template` (codegen fix),
  `packagingHandler.ts`.
- **Demo** (`dotnet/randomStub`): `Program.cs`, `randomStub.csproj`, `deploy.mjs`, `README.md`.

## Review order
Read commit-by-commit — the history is already sliced: channelization + wizard foundation → "Try it"
proof → GrammarGen reliability → build-in-scaffolder → isolated build → Try-it fix → randomStub demo.

## Follow-ups (not in this PR)
- **Make generated agents idiomatic workspace members.** Generated agents already land in
  `packages/agents/<name>` (the tracked repo, where discovery looks), but this PR builds them with an
  `--ignore-workspace` workaround and emits `link:` deps so the build never rewrites the shared lockfile
  or relinks native modules while the service is running. The follow-up makes the committed agent look
  like every hand-written agent — `workspace:*` deps and a relative tsconfig base — and has the wizard
  build it in place by linking its `@typeagent/*` deps to the already-built sibling packages (no
  workspace `pnpm install`, no lockfile churn). The lockfile delta is then recorded by the user's next
  ordinary install, like any added package.
- **Phase reconciliation against real outputs.** Validate that re-running an upstream phase marks
  downstream phases stale and that the diff preview reflects real artifact changes.
- **Conversational entry.** A thin `@typeagent create an agent for X` route that opens the wizard
  pre-seeded with the description.
- **End-to-end acceptance test.** A headless run (description → 7 phases → install → at least one
  utterance answered, health passes, no manual edits), with both a mocked-LLM and a live variant.


ts/packages/typeagent-studio/src/webviewKit/client/wizard.ts
```diff
@@ -0,0 +1,648 @@
...
+
+window.addEventListener("message", (event: MessageEvent) => {
+    const msg = event.data as HostToWizardMessage;
+    switch (msg.type) {
+        case "state":
+            model = msg.model;
+            errorText = undefined;
+            reconcileSelection();
+            render();
+            break;
```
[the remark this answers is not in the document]
TalZaccai: Addressed in 56a129f01. I avoided simply clearing `statusText` on every `state` message, because the banner is intentionally persistent for terminal messages the host posts right before a state refresh (e.g. "Started onboarding for X.", "Install cancelled."). Instead, the transient progress flows (`runPhase`, `rerunStale`, install-success) now post a terminal status, and the banner is cleared on `error`, so a stale "Running..." line no longer lingers.

ts/packages/agents/onboarding/src/lib/agentBuild.ts
```diff
@@ -0,0 +1,169 @@
...
+    let dir = path.resolve(startDir);
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
+        const npmrc = path.join(dir, ".npmrc");
+        if (existsSync(npmrc)) {
+            const match = readFileSync(npmrc, "utf-8").match(
+                /^\s*registry\s*=\s*(\S+)/m,
+            );
+            if (match) return match[1];
+        }
```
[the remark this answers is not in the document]
TalZaccai: Fixed in 56a129f01 — the `.npmrc` read is now best-effort (wrapped in try/catch); an unreadable file just keeps walking up the tree instead of crashing the build. Also dropped the now-redundant `existsSync` check, which removes the TOCTOU.

ts/packages/studio-service/src/onboarding/dispatcherGateway.ts
```diff
@@ -0,0 +1,281 @@
...
+/**
+ * Build (once) the onboarding-only dispatcher handle by lazily importing
+ * `default-agent-provider`. Throws an actionable error when the module can't be
+ * resolved (e.g. the packaged extension with `node_modules` stripped).
+ */
+async function getHandle(): Promise<OnboardingDispatcherHandleLike> {
+    if (handlePromise === undefined) {
+        handlePromise = (async () => {
+            let mod: DefaultAgentProviderOnboardingModule;
+            try {
```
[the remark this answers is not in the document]
TalZaccai: Good catch — fixed in 56a129f01. `getHandle()` now clears the memoized promise when the import/build attempt rejects, so a transient failure (e.g. a dev-time module-resolution hiccup) is retryable on the next call instead of wedging onboarding until the service restarts.

ts/packages/typeagent-studio/src/wizardView.ts
```diff
@@ -0,0 +1,365 @@
...
+    return vscode.workspace
+        .getConfiguration("typeagentStudio.onboarding")
+        .get<InstallHealthGatePolicy>("installHealthGatePolicy", "enforce");
+}
+
+/**
+ * Open (or reveal) the singleton New Agent wizard panel. Driven entirely by the
+ * in-process `runtime`; the panel keeps its context while hidden so navigating
+ * away and back doesn't drop the walked-phase state.
+ */
```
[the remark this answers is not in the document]
TalZaccai: Fixed in 56a129f01 — the doc comment now describes the panel as driven by the service-backed `OnboardingRuntime` rather than an in-process runtime.

ts/packages/typeagent-studio/src/webviewKit/wizardProtocol.ts
```diff
@@ -0,0 +1,138 @@
...
+ * Typed message envelope for the New Agent wizard's host (extension) ↔ webview
+ * (iframe) channel.
+ *
+ * Pure data + narrowing only — no `vscode` or DOM dependency, and only *type*
+ * imports from `@typeagent/core` — so both the extension host bundle and the
+ * browser client bundle import it, and it is unit testable. The host owns the
+ * in-process Studio runtime (onboarding is the one surface still driven in
+ * process) and turns each webview request into a runtime call, posting the
+ * resulting {@link WizardViewModel} back; the webview never touches the runtime
+ * directly.
```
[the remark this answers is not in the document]
TalZaccai: Fixed in 56a129f01 — the header no longer claims an in-process Studio runtime; it now says the host forwards each webview request to the standalone Studio service (where onboarding runs) and posts the resulting view model back.

dotnet/randomStub/deploy.mjs
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env node
```
robgruen: Are we sure this is the right place for this file?

robgruen: I think I see now what it's doing, I don't know if I like it though.

dotnet/randomStub/deploy.mjs
```diff
@@ -0,0 +1,157 @@
...
+function exeName() {
+    return process.platform === "win32" ? "randomstub.exe" : "randomstub";
+}
+
+// Directories we must never deploy into even if they look writable
+// (Windows reports false-positive write access on some protected stores).
+function isSystemDir(dir) {
+    const d = dir.toLowerCase();
+    return (
+        d.includes("\\windows\\") ||
```
robgruen: There are *many* more folders than this.  Instead of doing it like this why not include a list of approved folders instead (like only under the current users profile dir or something like that).

dotnet/randomStub/deploy.mjs
```diff
@@ -0,0 +1,157 @@
...
+
+    if (args.printTarget) {
+        console.log(targetDir);
+        return;
+    }
+
+    const publishDir = path.join(__dirname, "bin", "deploy", rid());
+    console.log(`Publishing randomStub (${rid()}, framework-dependent single-file)...`);
+    execFileSync(
+        "dotnet",
```
robgruen: what if dotnet's not installed?

dotnet/randomStub/Program.cs
```diff
@@ -0,0 +1,182 @@
+// Copyright (c) Microsoft Corporation.
```
robgruen: to address my weirdness from before why not pick a console app that exists on all three operating systems?

Image

ts/packages/agents/onboarding/src/grammarGen/grammarGenHandler.ts
```diff
@@ -433,6 +726,49 @@ function buildGrammarPrompt(
     ];
 }
 
+function buildGrammarRepairPrompt(
```
robgruen, line 756: TypeChat for LLM calls?

ts/packages/agents/onboarding/src/scaffolder/cliHandlerTemplate.ts
```diff
@@ -27,10 +27,33 @@ function flagToCamel(flag: string): string {
...
 }
 
-function buildSwitchCases(actions: DiscoveredAction[]): string {
+// The discovered `action.path` is the full command path including the base
+// command (e.g. "gh repo create" or "randomstub number"). The runtime handler
+// already invokes the base command as the executable via execFile, so the
+// switch cases must contribute only the subcommand tokens. Strip a leading
+// base-command prefix to avoid emitting it twice (which produced a bogus
+// `randomstub randomstub number ...` invocation).
+function subcommandTokens(fullPath: string, cliCommand: string): string[] {
```
robgruen, line 36: this looks like it might break if you do something like "gh repo create 'my repo'".

robgruen: shouldn't there be a command arg processing lib we can reuse here?

ts/packages/agents/onboarding/src/scaffolder/scaffolderHandler.ts
```diff
@@ -517,6 +543,51 @@ function buildMainSchemaWithSubGroups(
     };
 }
 
+// Remove a documentation comment attached directly (no blank line) above the
+// entry union `export type <Pascal>Actions = ...`. The action-schema compiler
+// (asc) treats such a comment as an "entry type comment" and errors out
+// because entry-type comments are not used for prompts. Comments separated by
+// a blank line, and comments on individual action types, are preserved.
+export function stripEntryTypeComment(
```
robgruen: This also seems brittle.  We should reuse the TS compiler or our SchemaParser to do this rather than trying to use regex.

ts/packages/agents/onboarding/src/scaffolder/scaffolderHandler.ts
```diff
@@ -789,21 +861,26 @@ function getWorkspaceDepValue(
...
     if (!pkgPath) return "workspace:*"; // unknown — best-effort fallback
+
     const absolute = path.join(TYPEAGENT_REPO_ROOT, pkgPath);
-    let rel = path.relative(targetDir, absolute).replace(/\\/g, "/");
+    const fromDir = path.resolve(targetDir ?? path.join(AGENTS_DIR, "_"));
+    let rel = path.relative(fromDir, absolute).replace(/\\/g, "/");
     if (!rel.startsWith(".")) rel = "./" + rel;
-    return `file:${rel}`;
+
+    // A freshly scaffolded agent is built in ISOLATION with
```
robgruen, line 828: LOVE THIS!

ts/packages/defaultAgentProvider/src/onboardingDispatcher.ts
```diff
@@ -0,0 +1,233 @@
...
+ * user's free-text agent description) stay in one token. The tokenizer treats a
+ * bare `'` as the token terminator and does NOT unescape `\'` on the way out
+ * (`stripQuoteFromToken`), so we cannot rely on backslash-escaping. Instead we
+ * make the payload contain no literal single quote at all: every `'` in the
+ * JSON only ever occurs inside a string value, and JSON permits it to be written
+ * as the `\u0027` escape, which `JSON.parse` decodes back to `'`. The result is
+ * both tokenizer-safe (no interior `'`) and valid JSON, for arbitrary text
+ * including apostrophes, double quotes, and backslashes.
+ */
+export function buildActionCommand(
```
robgruen, line 159: THis might be useful to have in the actual dispatcher

ts/packages/typeagent-core/src/runtime/studioRuntimeCore.ts
```diff
@@ -665,6 +665,46 @@ export interface AgentLocation {
     external: boolean;
 }
 
+/**
+ * Options for proving that a generated agent answers an utterance via
+ * "Try it". All optional: with none given, a deterministic PhraseGen example
+ * phrase is chosen and any resolved typed action counts as answered.
+ */
+export interface ProveUtteranceOptions {
```
robgruen, line 673: I would sync with Dominic, he's going to be creating the same interface so it would be good for you and him to discuss avoiding duplication.

ts/packages/typeagent-studio/src/webviewKit/client/wizard.ts
```diff
@@ -0,0 +1,648 @@
...
+let reconciliation:
+    | { restoredPhase: OnboardingPhaseName; stalePhases: OnboardingPhaseName[] }
+    | undefined;
+let utteranceProof: UtteranceProofResult | undefined;
+
+restorePersisted();
+
+// --- DOM helpers ---------------------------------------------------------
+
+function el(tag: string, className?: string, text?: string): HTMLElement {
```
robgruen: I would check the repo to see if this exists somewhere.  Seems like this and below are good candidates for an HTMLLib.

robgruen (approved):

dotnet/randomStub/deploy.mjs
TalZaccai: Agreed it didn't belong here. randomStub was only a demo fixture to exercise onboarding offline, so I've removed `dotnet/randomStub/` from the repo entirely (kept it out-of-repo for local demos/debugging). Resolving.

dotnet/randomStub/deploy.mjs
TalZaccai: Good point — the system-dir denylist was fragile and an allowlist would be safer. Rather than harden the deploy logic, I've dropped randomStub from the repo altogether since it was a demo-only tool. Resolving.

dotnet/randomStub/deploy.mjs
TalZaccai: Right, it would just crash. Since randomStub was only a demo fixture I've removed it from the repo rather than add a dotnet-presence guard. Resolving.

dotnet/randomStub/Program.cs
TalZaccai: Fair — and there isn't a truly universal, well-shaped, offline CLI across win/linux/osx, which is part of why I've removed the bespoke stub from the repo instead of swapping in another tool. It's kept out-of-repo for demos; a durable onboarding fixture can be a deliberate follow-up. Resolving.

ts/packages/agents/onboarding/src/scaffolder/cliHandlerTemplate.ts
TalZaccai: The generated handler runs via `execFile` (no shell), so parameter *values* are passed as discrete argv entries — `my repo` arrives as a single argument, spaces intact, with no quoting/escaping and no shell-injection risk. `subcommandTokens` only splits the discovered *command path* (subcommand names like `gh repo create`) to strip the base-command prefix; it never touches user-supplied values. Arg-parsing libs (yargs/commander/minimist) go the other direction — turning an argv string into options — whereas here we emit argv from already-typed parameters into an `execFile` array, which is the safe minimal form. Resolving, but happy to revisit if you had a specific lib in mind.

ts/packages/agents/onboarding/src/scaffolder/scaffolderHandler.ts
TalZaccai: Thanks! This `link:`/`file:` resolution was fiddly to get right for the isolated (`--ignore-workspace`) agent build, glad it reads clearly.

ts/packages/defaultAgentProvider/src/onboardingDispatcher.ts
TalZaccai: Agreed. `buildActionCommand` exists only because the dispatcher's sole entry point is the `@action` text command (there's no structured-action API), which is why it has to encode the JSON tokenizer-safely (`\u0027`). If the dispatcher grows a structured-action entry point or a shared command-builder, this is a good candidate to promote there. Nothing to change in this PR, so resolving as a forward-looking note.

ts/packages/agents/onboarding/src/grammarGen/grammarGenHandler.ts
TalZaccai: The grammar generation uses a raw `model.complete` on purpose: the meaningful validation here isn't JSON shape but whether the produced `.agr` actually *compiles*, so the loop feeds agc's exact compiler error back for a targeted repair pass (up to 4 attempts). TypeChat's `createJsonTranslator` would validate/repair the response *shape*, but the envelope is a trivial `{ grammar: string }` — it can't judge grammar compilation, so it wouldn't replace the compile→repair loop. I'm happy to route the raw call through our standard LLM wrapper for consistency as a follow-up, but keeping the agc-driven repair loop is the substantive part. Leaving open for your call.

ts/packages/agents/onboarding/src/scaffolder/scaffolderHandler.ts
TalZaccai: Agreed that regex over TS source is inherently fragile. This one was deliberately narrow — it strips a comment block only when it sits *immediately* above the entry union and stops at the first blank or code line — but a TS-compiler/SchemaParser approach (locate the union node, use `getLeadingCommentRanges`) is more robust. Addressed in ae7dedc2a — details below.

ts/packages/typeagent-core/src/runtime/studioRuntimeCore.ts
TalZaccai: Good call — I'll sync with Dominic before we finalize these interfaces so we don't end up with two parallel definitions. Leaving this open as the tracking point until we've aligned.

ts/packages/typeagent-studio/src/webviewKit/client/wizard.ts
TalZaccai: Confirmed — there was real duplication: `el` was defined privately here and in `impactReport.ts`, and already *exported* from `traceViewerDom.ts` alongside `clear`. A small shared `domHelpers.ts` under `webviewKit/client` (exporting `el`/`clear`/`button`) that all three import removes the triplication. Addressed in ae7dedc2a — details below.

ts/packages/agents/onboarding/src/scaffolder/scaffolderHandler.ts
TalZaccai: Done in ae7dedc2a. Rewrote `stripEntryTypeComment` to use the TypeScript parser instead of regex over the source text: `ts.createSourceFile` locates the `export type <Pascal>Actions` alias, then `ts.getLeadingCommentRanges` walks the leading comments and strips only the block directly attached to the declaration (no blank line between them). Comments separated by a blank line and per-action comments are preserved, and it's a no-op when there's no matching union. `typescript` is promoted to a runtime dependency to match the actionSchema/typeagent convention for importing the compiler API.

ts/packages/typeagent-studio/src/webviewKit/client/wizard.ts
TalZaccai: Good call — `el`/`clear` already lived in `traceViewerDom.ts`. Consolidated all three (`el`, `clear`, `button`) into a shared `webviewKit/client/domHelpers.ts` in ae7dedc2a. `wizard.ts` and `impactReport.ts` now import from it, and `traceViewerDom.ts` re-exports `el`/`clear` from the shared module so its render modules are untouched. The shared `el(tag, className?, text?)` is a superset of the old 2-arg version, so existing call sites are unaffected.

ts/packages/agents/onboarding/src/grammarGen/grammarGenHandler.ts
robgruen: Maybe using TypeChat is heavy handed, BUT, we can use TypeChat methods here.  It doesn't necessarily have to output JSON but you could do something like: {
  grammar: string;
  compilationResult: boolean;
  error?: string
}

which *might* actually help.  I'd want to experiment but I think you could probably get slightly faster repairs if you use some structure.  It's fine leaving as is but maybe put a comment in saying why we chose this path over the typechat path.

robgruen:
What if you had a schema that contains a string that is itself a schema that matches this pattern?

robgruen:
Curtis write a schema parser that's entirely self-contained in our repo. 🙂  No need for the dependency.

ts/packages/agents/onboarding/src/grammarGen/grammarGenHandler.ts
TalZaccai: Done in d2a6181c9. Routed the grammar-gen call through TypeChat (`createTypeScriptJsonValidator` + `createJsonTranslator`) so it matches the codebase's structured-LLM convention. The prompts already carry the full instructions and the `{ grammar }` JSON contract, so `createRequestPrompt` is suppressed and the call is driven from our own prompt sections (same pattern as the markdown agent). TypeChat now handles typed extraction/validation of the envelope plus a shape-repair round, so the manual `extractGrammarContent` salvage is removed. The substantive part — the agc compile→repair loop — is unchanged.

ts/packages/typeagent-core/src/runtime/studioRuntimeCore.ts
TalZaccai: Synced with Dominic on this — we're aligning on a single shared interface rather than two parallel definitions, so there's no duplication to reconcile in this PR. Tracking the concrete shape as we converge outside this thread. Marking as completed.

TalZaccai:
> What if you had a schema that contains a string that is itself a schema that matches this pattern?

Handled — this is exactly why moving off the regex to the TS parser matters. A string literal's contents are never parsed as a `TypeAliasDeclaration`, so `getLeadingCommentRanges` on the real entry-union node can't be fooled by schema-looking text inside a string. Only the comment attached directly above the actual `export type <Name>Actions` union is stripped.

Verified in 44f918824 with a test covering precisely that: a `{ s: "export type FooActions = X;" }` field is left untouched while the real entry-type comment is stripped.

TalZaccai:
> Curtis wrote a schema parser that's entirely self-contained in our repo. 🙂 No need for the dependency.

Agreed — done in 44f918824. Moved `stripEntryTypeComment` into `@typeagent/action-schema` (it's really an action-schema concern: the parser there is what rejects entry-type comments in the first place) and exported it from the package. Onboarding now imports the shared helper and I dropped its direct `typescript` dependency. So we reuse the in-repo schema tooling rather than reaching for the compiler API directly.

PR #2760, opened 2026-07-30
Onboarding: Generate real REST handlers for OpenAPI-discovered agents

TalZaccai:
## Summary

When the onboarding pipeline discovers an integration from an OpenAPI spec, the
scaffolder previously emitted only a **stub** action handler: the generated
agent could translate an utterance into an action, but the handler never called
the API. This PR adds a deterministic, `fetch`-based REST handler generator so
an OpenAPI-discovered agent issues real HTTP requests end to end.

## What changed

- **Discovery (`parseOpenApiSpec`)** now captures the two things a REST handler
  needs:
  - **Base URL** — resolved from `servers[0].url`, with server-variable
    substitution and relative-URL resolution against the spec's fetched
    location; falls back to the spec origin when there is no `servers` entry.
    Left unset when it can't resolve to a well-formed http/https URL.
  - **Parameter location** — each parameter records its wire `in`
    (path/query/header/cookie, or `body` for request-body fields). Path-level
    and operation-level parameters are merged, and local
    `#/components/parameters/*` `$ref`s are resolved inline.
  - Both are persisted on the `ApiSurface` and survive approval.

- **Scaffolder** routes to the new REST generator when the surface has a
  resolved `baseUrl` and REST-shaped actions; otherwise it falls back to the
  existing stub. Explicit non-REST patterns (websocket-bridge, native-platform)
  keep their own builders.

- **Generated handler** (from `restHandler.template`) builds the request URL
  from path/query parameters, sends a JSON body for POST/PUT/PATCH, applies a
  30s timeout, and returns truncated/pretty-printed responses with HTTP errors
  surfaced.

## Scope / v1 limitations

- JSON OpenAPI 3 specs only (Swagger 2 and YAML yield no actions).
- Scalar path/query params only (no arrays/objects/style/explode).
- Header/cookie params: required ones fail the action, optional ones are
  dropped. No auth headers. DELETE requests carry no body.
- Non-local `$ref`s (external files, deeper pointers) are skipped; an action
  whose path placeholder can't be resolved falls back to the stub handler.

## Testing

Adds a `tsx --test` runner and 30 unit/integration tests: OpenAPI action
extraction, base-URL resolution, api-surface preservation, scaffolder REST
routing, codegen fixes, generated-handler compilation, and live
execution/E2E against the Open Library `/api/books` endpoint. All 30 pass.


ts/packages/agents/onboarding/src/scaffolder/restHandlerTemplate.ts
```diff
@@ -0,0 +1,235 @@
...
+// only — crawlDocUrl's LLM-guessed method/path is out of scope).
+//
+// v1 limitations (see restHandler.template header comment too):
+//   - scalar path/query params only (no arrays/objects, no style/explode)
+//   - header/cookie params are unsupported: required ones fail the action,
+//     optional ones are silently dropped
+//   - no auth (no Authorization / api-key headers)
+//   - DELETE requests never carry a body
+//   - `$ref`-based params/bodies/path-items are skipped upstream in
+//     discoveryHandler.ts, so they never reach this generator
```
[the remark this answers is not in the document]
TalZaccai: Good catch — this is already handled on the branch. The misleading `$ref` 'skipped upstream' bullet has been removed from the restHandlerTemplate.ts v1-limitations comment, and `extractOpenApiActions` now resolves local `#/components/parameters/*` `$ref`s inline (via `resolveLocalParameterRef`) so referenced params aren't dropped. Only non-local/unsupported `$ref`s (external files, deeper pointers, `$ref`'d bodies/path-items) are skipped. Resolving.

ts/packages/agents/onboarding/src/scaffolder/restHandler.template
```diff
@@ -0,0 +1,105 @@
...
+    AppAgent,
+    TypeAgentAction,
+} from "@typeagent/agent-sdk";
+import {
+    createActionResultFromTextDisplay,
+} from "@typeagent/agent-sdk/helpers/action";
+import { {{PASCAL_NAME}}Actions } from "./{{NAME}}Schema.js";
+
+const BASE_URL = {{BASE_URL}};
+const MAX_RESPONSE_CHARS = 4000;
```
robgruen: Mabe move this to the aiclient and maybe spec it out based on model...right now a lot of these are just arbitrary and don't really match the selected mode.  We should have a function or config that sets this that callers can use instead of embedding this constant all over the place.

ts/packages/agents/onboarding/src/scaffolder/restHandler.template
```diff
@@ -0,0 +1,105 @@
...
+
+function truncate(text: string): string {
```
robgruen: we might already have this function in textprocessing lib

robgruen (approved):

ts/packages/agents/onboarding/src/scaffolder/restHandler.template
TalZaccai: Good call — there isn't actually a reusable truncate in a shared lib (textpro only has a private `simplifyText`; the other `truncate*` fns are app-specific). Since the generated handler is standalone code that already depends on `@typeagent/agent-sdk` (but not textpro/aiclient), I added a reusable `truncateText(text, maxChars?)` to agent-sdk's action helpers and the template now imports it instead of inlining its own. (8503913d58426d8094f0cfb3b91da7438452e341)

ts/packages/agents/onboarding/src/scaffolder/restHandler.template
TalZaccai: Agreed the magic 4000 shouldn't be embedded. Pulled it into a single named `DEFAULT_MAX_RESPONSE_CHARS` in agent-sdk (next to the shared `truncateText`) so it lives in one place and callers reference it. Full model-aware sizing (budget derived from the selected model) really belongs in the host/dispatcher truncating tool results — the generated REST handler has no model context — so I've tracked that as a follow-up rather than bake fake model-awareness into standalone codegen. Happy to pick it up there if you'd like it in this PR. (8503913d58426d8094f0cfb3b91da7438452e341)

PR #2764, opened 2026-07-30
fix: correct list/weather agent schema routing and grammar

jebrans:
# Summary

Improve list/weather agent routing and prevent over-generalized
construction cache entries from hijacking unrelated requests.

- Add explicit schema guidance and grammar coverage for list actions
  (`createList`, `getList`, `clearList`, `deleteList`, `listLists`)
  to improve intent selection and disambiguation.
- Add support for deleting lists through a new `deleteList` action.
- Improve weather action selection by adding descriptive schema
  guidance and an explicit "what is the weather" trigger phrase.
- Reject anchorless learned constructions in the construction cache
  to prevent broad wildcard rules from matching nearly any input.
- Add regression tests covering anchorless construction rejection
  and anchored construction acceptance.

# Example

```text
What is the weather?

→ Routed to list.getList
→ Agent searches for a list named "weather"

What is the weather?

→ Routed to weather.getCurrentConditions
```


ts/packages/agents/list/src/listActionHandler.ts
```diff
@@ -453,6 +457,21 @@ async function handleListAction(
             result.entities = getEntities(listName);
             break;
         }
+        case "deleteList": {
```
robgruen, line 471: Maybe inject a user yes/no confirm request here.  Or better yet maybe for a demo we could just tombstone it so the user could say "can you undelete/restore the list we just deleted"

ts/packages/agents/list/src/listSchema.ts
```diff
@@ -7,21 +7,29 @@ export type ListAction =
...
 export type ListActivity = StartEditList;
 
+// Name of a list such as "grocery", "to do", "shopping", "packing", "gift",
+// "book", "idea", "movie", "garden task", "place to visit". Names should be
+// lower case and stemmed to the singular form (e.g., "movies" -> "movie").
+// IMPORTANT: Do not invent a list name. If the user uses a reference phrase
+// ("the list", "that list", etc.), clarify with the user which list they
+// meant unless it is obvious from the conversation history. If the request
+// does not name an actual list (or is unrelated to lists altogether), use
+// the "unknown" action instead of guessing a listName.
```
robgruen: Actually it should do "clarify* actions.  We shouldn't bials the model here so we should take this out if we can.

ts/packages/agents/weather/src/weatherSchema.ts
```diff
@@ -14,6 +19,11 @@ export type GetCurrentConditionsAction = {
     };
 };
 
+// get the weather forecast (optionally for a number of days) for a location,
+// for example "what's the forecast", "weather forecast for tomorrow",
+// "forecast for the next 5 days in Chicago". Requires a location — if the
+// user doesn't name one, ask them which location they mean instead of
+// guessing.
```
robgruen: I believe the users location if known is provided in the translation prompt. Another option is for Location to be optional and if in the action handler there is no location we can ask the user then.  We probably want to remove explicit directives like this from the schema.  Can we try it and see what happens?

ts/packages/agents/list/src/listActionHandler.ts
jebrans: Will do, let me work on soft delete with ability to restore

ts/packages/agents/list/src/listSchema.ts
jebrans: Will remove

hillary-mutisya (approved):

PR #2769, opened 2026-07-31
Recover translation-bench checkpoints after interrupted writes

datduyng:
Makes single-writer translation-bench checkpoints recover from interrupted JSONL writes while preserving validated resume and merge behavior.

```console
Before: header\n{"caseId":"1"}\n{"caseId":
Append: {"caseId":"2"}
After:  header\n{"caseId":"1"}\n{"caseId":"2"}
Test:   PASS translationBench.checkpointPrimitives.spec.js (1 passed, 0 failed)
```


ts/packages/benchmarks/src/translationBench/synthesizer/jsonlCheckpoint.ts
```diff
@@ -0,0 +1,104 @@
...
+ * Concurrent calls for the same path are not supported.
+ */
+export function appendSyncedJsonlRecords(
+    filePath: string,
+    records: readonly string[],
+): void {
+    if (records.length === 0) return;
+    const handle = fs.openSync(filePath, "r+");
+    try {
+        const content = fs.readFileSync(handle);
```
jebrans, line 69: We load the entire file to read the last line, Since we parsed it earlier can we use that to retrieve the last line

ts/packages/benchmarks/src/translationBench/synthesizer/generationSupport.ts
```diff
@@ -339,6 +338,58 @@ export function appendTranslationBenchCheckpointRows<T = unknown>(
     };
 }
 
+export function mergeTranslationBenchCheckpoints<T = unknown>(
```
jebrans: Are we using this function?

datduyng: This pr is a subset of https://github.com/microsoft/TypeAgent/pull/2831 - let me add the usage in this pr as well

ts/packages/benchmarks/src/translationBench/synthesizer/jsonlCheckpoint.ts
datduyng: I have experienced many time that mid-run there is a crash due to excessive io and memory usage on my computer causing broken json line writes. the whole point of a read all and write is to make it safe atomic. At this point, I waste lot of time on failed run thus - just making the system more robust for local run. I am hesitant on pulling in many external libraries until I know for sure they are well maintained.

Also this code only run local and no prod impact.

In the future, I will consider pulling in a jsonl library processor to manage this but right now, I don't want overbloat the code with many libraries that are unmaintained.

datduyng:
```
try {
    JSON.parse(tail);
    fs.appendFileSync(filePath, "\n");
} catch {
    fs.truncateSync(filePath, lastNewline + 1);
}
```

if a crash has

```
{"case":"1"}
{"case":
```

json parse would failed and it get removed -

jebrans (approved):

ts/packages/benchmarks/src/translationBench/synthesizer/jsonlCheckpoint.ts
datduyng: Fixed in 025b7dab33336ea3819cb1f927164043f18fb432

PR #2771, opened 2026-07-31
[translation-bench] Part 1/n: scaffold + generate price snapshots + catalogs

datduyng:
Part 1 of the translation-bench series: package scaffold and pinned generators only (no runner/matrix yet).


ts/packages/benchmarks/src/core/paths.ts
```diff
@@ -0,0 +1,34 @@
...
+// so ../.. is the package root.
+const packageRoot = path.join("..", "..");
+
+/** Resolve a path relative to the package root (works from compiled dist/). */
+export function getPackageFilePath(packageRootRelativePath: string): string {
+    if (path.isAbsolute(packageRootRelativePath)) {
+        return packageRootRelativePath;
+    }
+    return fileURLToPath(
+        new URL(
```
jebrans, line 17: Will this work with backward slashes

datduyng: `import path from "node:path";` handle both macos and window path

jebrans (approved):

ts/packages/benchmarks/src/core/paths.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+// Same pattern as cache/defaultAgentProvider: this file emits to dist/core/,
+// so ../.. is the package root.
+const packageRoot = path.join("..", "..");
+
+/** Resolve a path relative to the package root (works from compiled dist/). */
+export function getPackageFilePath(packageRootRelativePath: string): string {
```
robgruen, line 12: We have this function in several places.  We should consolidate them into the common or utils packages at some point.

ts/packages/benchmarks/src/core/model-prices.generated.json
```diff
@@ -0,0 +1,58 @@
+{
```
robgruen, line 1: Maybe generate this file on the fly based on which models we are benchmarking.

PR #2775, opened 2026-07-31
fix bare <Rule>? grammars to (<Rule>)? + quantifier RFC

datduyng:
- Rewrite all bare `<Name>?` sites in shipping/test `.agr` files to the supported `(<Name>)?` form so optional rule refs actually match.

On main, `?` after a rule ref is a **literal** token. Quantifiers only apply after `)`. Authors wrote `<Polite>?` meaning optional; those patterns never matched.


curtisman (approved):

curtisman:
The description mention an RFC,  which is not in this PR,  right?

PR #2784, opened 2026-08-01
Match only named lists in list grammar

datduyng:
Addresses review feedback (Robert): fix bare-reference capture by **removing over-broad list grammar rules**, not by teaching the list agent about placeholders/caching.

**Grammar matches (named lists only)**
- `add milk to the grocery list` / `add milk to my shopping list`
- `can you add milk and eggs to the grocery list`
- `put cheese on the grocery list`, remove/get/clear with the same named shape

**Grammar does not match → translation (LLM + history)**
- `add ham to the list`, `put cheese on the list`, `add cheese to both`, `add ham to my list`
- Also determinerless `add milk to grocery list` (avoids reopening `listName="the"` via `to $(listName) list`)


robgruen:
Looking at this I think a better approach is to just remove the grammar rule that's causing the problem.

If I say: "create a ingedients list and a shopping list" and then "add cheese to both" that should work. The agent shouldn't know or care about any grammar or caching.  In fact, it should work regardless of language. A user should be able to create a list called "my"... or, my could refer to a list of the useres name but that should get resolved in translation and not in the agent.

datduyng:
@robgruen ack! Thank you!

GeorgeNgMsft (approved):

PR #2796, opened 2026-08-04
Parse Seal-Tools source conversations

datduyng:
Parses Seal-Tools conversation rows, Python-literal calls, and per-row function schemas for translation benchmarks.

```console
$ pnpm run jest-esm --runInBand --testPathPattern=translationBench.sealToolsParser.spec.js
Result: 1 suite passed, 1 test passed, 0 failed
$ node --input-type=module -e "import('./dist/translationBench/public_datasets/Seal-Tools/getDataset.js').then(...)"
Result: Find trips, search
```


ts/packages/benchmarks/src/translationBench/public_datasets/Seal-Tools/getDataset.ts
```diff
@@ -52,3 +58,233 @@ if (process.argv[1] === fileURLToPath(import.meta.url)) {
...
+        offset: catalogIndex + API_LIST_MARKER.length,
+    });
+    if (!Array.isArray(catalog.value)) {
+        throw new Error("human turn API catalog must be an array");
+    }
+
+    const taskIndex = value.indexOf(TASK_MARKER, catalog.end);
+    if (taskIndex < 0) throw new Error("human turn has no task instruction");
+    const instructionStart = taskIndex + TASK_MARKER.length;
+    const parsedInstruction = parsePythonLiteralAt(value, {
```
GeorgeNgMsft: Apparently, parsePythonLiteralAt returns immediately upon encountering the next matching quote character, which is normally fine. But apparently 1/7 of the seal corpus rows includes tasks strings with unescaped matching quotes, which will cause the instructions to be truncated unintentionally.

Consider identifying the proper first/last delimiters to properly isolate the full instruction text.

ts/packages/benchmarks/src/translationBench/public_datasets/Seal-Tools/getDataset.ts
```diff
@@ -52,3 +58,233 @@ if (process.argv[1] === fileURLToPath(import.meta.url)) {
...
+    };
+}
+
+// Parse the assistant's Python-repr list while preserving number lexemes for
+// exact corpus scoring.
+function parseAssistantTurn(value: string): SealToolsCall[] {
+    const trimmed = value.trim();
+    if (trimmed === "" || trimmed === "-1") return [];
+    const parsed = parsePythonLiteral(trimmed, {
+        preserveNumberLexemes: true,
```
GeorgeNgMsft, line 223: Apparently this gets convereted into 
{
    parameters: {
        age: new PythonNumber("N"),
    },
}

which in sealToolsScorer gets compared like so

```
// sealToolsScorer.ts:124-138
if (typeof value === "number") return String(value);

if (typeof value === "object") {
    return `{${Object.entries(value)
        .map(([key, item]) => `${pythonRepr(key)}: ${pythonRepr(item)}`)
        .join(", ")}}`;
}
```
but there is a risk here where sealToolsScorer.ts does not handle PythonNumber and that can result in this comparison 

"44" === "{'lexeme': '44'}" 

You may need to find a way to fix the comparison and maybe add test coverage for this too.

GeorgeNgMsft (approved):
Approved with a few follow up items to address before merge

ts/packages/benchmarks/src/translationBench/public_datasets/Seal-Tools/getDataset.ts
datduyng: Fixed in 401312695. PythonNumber now compares by its source lexeme, including the case-insensitive path. Added regression coverage.

ts/packages/benchmarks/src/translationBench/public_datasets/Seal-Tools/getDataset.ts
datduyng: Fixed in 401312695. The parser now uses the first and final task delimiters. Verified all 700 rows with zero mismatches.

PR #2806, opened 2026-08-06
[OTEL] Add OpenTelemetry provider lifecycle management

GeorgeNgMsft:
**Adds lifecycle coordination for OpenTelemetry in TypeAgent-owned processes.** Partner hosts are expected to manage their own OpenTelemetry providers and lifecycle (we may offer a partner bootstrap later on).

 - Adds idempotent  initTelemetry()  and  shutdownTelemetry()  APIs.
 - Creates only the requested trace, metric, and log providers.
 - Shares resource information across all providers.
 - Prevents replacing providers installed by another owner.
 - Supports injected providers and components for testing.
 - Flushes and shuts down owned telemetry components with bounded timeouts.
 - Handles partial initialization failures and repeated shutdown calls safely.
 
 Depends on : https://github.com/microsoft/TypeAgent/pull/2801


ts/packages/telemetry/src/otel/redaction.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Recursively apply `redact` to every string in `value`, preserving its
+ * shape. Only secret *detection* lives in `@typeagent/common-utils`
+ * (`filterSecrets`); this walk exists solely to let a stateful
+ * {@link SecretFilter} - which only exposes `filter(text)`, not its
+ * underlying values - participate in structural redaction the same way
+ * `filterSecretsFromObject` does for the stateless case above.
+ */
+function walkStrings(
```
datduyng: ts/packages/utils/commonUtils/src/secretFilter.ts - just run bunch of regexes. 

This recursive walk is very expensive and slow. One regex on the whole body is enough. Nested walk of regexes could be expensive here

ts/packages/telemetry/src/otel/lifecycle.ts
```diff
@@ -0,0 +1,204 @@
...
+    readonly onShutdown: ShutdownCallback;
+}
+
+/** Create a new, empty {@link TelemetryLifecycle}. */
+export function createTelemetryLifecycle(
+    options: TelemetryLifecycleOptions = {},
+): TelemetryLifecycle {
+    const totalTimeoutMs = options.totalTimeoutMs ?? 10_000;
+    const componentTimeoutMs = options.componentTimeoutMs ?? 5_000;
+    if (!Number.isFinite(totalTimeoutMs) || totalTimeoutMs <= 0) {
```
datduyng: according to codex,

`setTimeout` does not support every positive finite number: non-integer delays are truncated and Node clamps delays greater than 2,147,483,647 ms to a near-immediate timeout. Consequently, a caller can supply a value accepted here but get an immediate or shorter-than-requested shutdown failure. Require an integer within Node's supported timer range for both timeout options.

    if (!Number.isFinite(totalTimeoutMs) || totalTimeoutMs <= 0) {
        throw new Error(
            "Telemetry lifecycle totalTimeoutMs must be a positive finite number.",
        );
    }
    if (!Number.isFinite(componentTimeoutMs) || componentTimeoutMs <= 0) {
        throw new Error(
            "Telemetry lifecycle componentTimeoutMs must be a positive finite number.",
        );
    }
Suggestion

    const maxTimerMs = 2_147_483_647;
    if (
        !Number.isInteger(totalTimeoutMs) ||
        totalTimeoutMs <= 0 ||
        totalTimeoutMs > maxTimerMs
    ) {
        throw new Error(
            "Telemetry lifecycle totalTimeoutMs must be a positive integer within the supported timer range.",
        );
    }
    if (
        !Number.isInteger(componentTimeoutMs) ||
        componentTimeoutMs <= 0 ||
        componentTimeoutMs > maxTimerMs
    ) {
        throw new Error(
            "Telemetry lifecycle componentTimeoutMs must be a positive integer within the supported timer range.",
        );
    }

ts/packages/telemetry/src/otel/lifecycle.ts
```diff
@@ -0,0 +1,204 @@
...
+    async function runShutdown(): Promise<void> {
+        const failures: unknown[] = [];
+        const deadline = Date.now() + totalTimeoutMs;
+        // Reverse registration order: the most recently registered
+        // component (typically the one most dependent on earlier ones)
+        // shuts down first.
+        for (let i = components.length - 1; i >= 0; i--) {
+            const component = components[i];
+            const remainingMs = deadline - Date.now();
+            if (remainingMs <= 0) {
```
datduyng, line 126: Another one from copilot,

```
Once the shared deadline is consumed, this skips each earlier-registered component entirely. In the coordinator, those include provider shutdown callbacks, so a slow later callback can leave a provider or custom writer without any shutdown attempt. This also conflicts with the public contract that `shutdown()` shuts down every registered component. Preserve the total wall-clock bound only if callbacks can be cancelled; otherwise invoke the remaining callbacks (with an explicit best-effort policy) rather than silently skipping them.

            if (remainingMs <= 0) {
                failures.push(
                    new TelemetryShutdownTimeoutError(component.name, 0),
                );
                continue;
            }
```

datduyng (approved):

ts/packages/telemetry/src/otel/redaction.ts
GeorgeNgMsft: good call

PR #2808, opened 2026-08-06
part 2/n catalog generator and action parameter deterministic grader

datduyng:
## Summary
- Add action-parameters grader generation (`genActionParametersGrader`) and check in `action-parameters-grader.generated.json` (557 actions).
- Fix catalog `type-union` fidelity so moniker|string fields (e.g. `browser.openWebPage.site`) become open `string` with free_text/nonempty scoring instead of opaque `any`/ignore.
- Make `genCatalog` schema-only (no full agent runtime) to avoid OOM; regenerate `catalog.generated.json`.

## Test plan
- [x] `cd ts/packages/benchmarks && pnpm run build`
- [x] `pnpm run jest-esm --testPathPattern=catalogGenerator` (40 tests)
- [x] `node dist/translationBench/scripts/genCatalog.js` → 557 actions, 0 bare `type-union`
- [x] `node dist/translationBench/scripts/genActionParametersGrader.js` → 557 actions, site free_text/nonempty

## Notes
- Regenerate with `pnpm run gen-catalog` in `@typeagent/benchmarks`.
- LLM fallback optional via `--model` when regex misses; current artifact is all-regex.


ts/packages/benchmarks/src/translationBench/scripts/genCatalog.ts
```diff
@@ -19,212 +40,672 @@ interface GeneratedAction {
...
 
-    const { provider, schemaNames } = await getAllActionConfigProvider(
-        getDefaultAppAgentProviders(undefined),
+/** Built-in dispatcher + system manifests (paths relative to dispatcher package). */
+function builtinManifests(
+    dispatcherRoot: string,
+): Record<string, AppAgentManifest> {
+    const resolve = (rel: string) => path.resolve(dispatcherRoot, rel);
+    return {
+        dispatcher: {
```
GeorgeNgMsft, line 262: This already exists elsewhere right? We shouldn't hardcode it, better to extract it into a small, shared module (if the existing one is too large) and just export/import where needed.

GeorgeNgMsft (approved):

ts/packages/benchmarks/src/translationBench/scripts/genCatalog.ts
datduyng: I agree overall. in this case, it's for benchmark purposes thus I need a snapshot as source code can always changes. that's why if you look at many places in the benchmark I am generating *.generated.json snapshots with a hash of the source code.

PR #2811, opened 2026-08-06
Add pinned Hugging Face dataset downloads

datduyng:
Adds revision-pinned Hugging Face row downloads with validated pagination, bounded retries, cancellation, and atomic JSONL replacement.

```console
Source: owner/name@abc123, split=train
Pages:  offset=0 -> {"id":1}; offset=1 -> {"id":2}
Output: {"id":1}\n{"id":2}
Test:   PASS translationBench.huggingFaceRows.spec.js (1 passed, 0 failed)
```

PR #2813, opened 2026-08-06
part 3/n simple-action dataset synthesizer

datduyng:
CleanShot 2026-08-06 at 15 17 58@2x

- Add simple-action dataset synthesizer for translation-bench (one expected action per positive/seed).
- Versioned Zod parsing for JSONL rows (`metadata`/`case` v1) and generation checkpoints so later formats can register new schema versions.
- Source adapter seam + seed-qa adapter, quality verifier, generation schedule/checkpoint/resume.
- `actionShape` policy defaults to `simple`; `multi` is reserved for a follow-up PR without rewriting the dataset contract.
- No runner or dispatcher commands yet (later parts).

**Stack / base**
- Targets **`main`** (reopened off the previous stacked draft #2811).
- Includes part 2 commits until #2808 lands; once #2808 merges, this diff shrinks to synthesizer-only.
- Part 1 already on main (#2771). Part 2: #2808 (approved).


ts/packages/benchmarks/src/translationBench/synthesizer/adapters/seedQaJsonlAdapter.ts
```diff
@@ -0,0 +1,379 @@
...
+    sourceText: string,
+    options: TranslationBenchSourceImportOptions,
+): TranslationBenchSourceCandidate[] {
+    validateManifest(options.manifest);
+    if (sha256Text(sourceText) !== options.manifest.sourceFileHash) {
+        throw new Error(
+            "source file hash does not match the pinned manifest",
+        );
+    }
+    const rows = parseJsonl(sourceText);
```
GeorgeNgMsft, line 354: Malformed json can break your import here (regardless of the skipInvalidRows option) as it happens before the guards take effect.

Copilot suggests parseJsonl() just returning the raw JSONL lines and doing the selected line parsing inside the per-row try below so that it can properly skip.

**Suggestion :** 

Instead,  parseJsonl()  should return raw JSONL lines, and each selected line should be parsed and validated inside the existing per-row  try :
```
try {
    const rawRow = parseSeedQaRow(
        parseJsonText(lines[rowIndex], `seed-qa line ${rowIndex + 1}`),
        `seed-qa line ${rowIndex + 1}`,
    );
    // Import candidate...
} catch (error) {
    if (!options.skipInvalidRows) throw error;
}
```

ts/packages/benchmarks/src/translationBench/synthesizer/generationSupport.ts
```diff
@@ -0,0 +1,430 @@
...
+        return parsed;
+    });
+
+    let current: TranslationBenchCheckpoint<T>;
+    if (fs.existsSync(filePath)) {
+        current = readTranslationBenchCheckpoint<T>(filePath);
+        assertCompatibleHeaders(current.header, header);
+    } else {
+        try {
+            fs.writeFileSync(filePath, `${canonicalJson(header)}\n`, {
```
GeorgeNgMsft, line 303: There's a niche scenario where the write could be interrupted for one reason or another, and the \n newline character doesn't get appended. This will leave you a line fragment, which seems to trigger an exception below and will prevent your checkpoint from resuming. 

Since this is pretty easily identified (missing \n), you should consider handling that case directly and repeat the row if needed, so that the checkpointing will be more robust.

ts/packages/benchmarks/src/translationBench/synthesizer/adapters/seedQaJsonlAdapter.ts
datduyng: this is expected. one jsonl line failure should crash the whole system in this case. the "skipInvalidRows" is more for contract failures but syntax errors should crash the system. Let me know if you have any other questions.

ts/packages/benchmarks/src/translationBench/synthesizer/generationSupport.ts
datduyng: Discussed offline. This is expected behavior

GeorgeNgMsft (approved):

PR #2820, opened 2026-08-07
part 4/n - flag parameter that required to graded through llm as a judge

datduyng:
Soft-match NL dates and add dataset audit so the action-parameters grader is safe at 1k scale.


ts/packages/benchmarks/src/translationBench/synthesizer/datasetGenerator.ts
```diff
@@ -258,7 +312,7 @@ export function createTranslationBenchGenerationSchedule(
             schemaCount: census.schemaCount,
             actionCount: census.actionCount,
             scheduledActionCount,
-            complete: scheduledActionCount === census.actionCount,
+            complete: scheduledActionCount === eligibleActionCount,
```
GeorgeNgMsft, line 299: The datasetGenerator now considers complete defined as _scheduledActionCount === eligibleActionCount_ which makes sense. But it seems the benchmark validation still has an outdated definition which may result in excluded actions failing validation

Image

ts/packages/benchmarks/src/translationBench/synthesizer/catalogGenerator/actionParametersGrader.ts
```diff
@@ -342,10 +504,12 @@ function classifyStringFieldRegex(
         };
     }
     if (isDateName(name)) {
+        // NL relative dates dominate synthesis ("next Tuesday", "this week").
+        // Exact string match is unfair at eval; align with time → nonempty.
         return {
             create: "temporal",
-            verify: "exact",
-            rule: "string-date-exact",
+            verify: "nonempty",
```
GeorgeNgMsft, line 511: Won't nonempty allow any value, including invalid ones? Can this be changed to "temporal"?

GeorgeNgMsft (changes requested):
Left a couple comments above

ts/packages/benchmarks/src/translationBench/synthesizer/datasetGenerator.ts
datduyng: resolved!

ts/packages/benchmarks/src/translationBench/synthesizer/catalogGenerator/actionParametersGrader.ts
GeorgeNgMsft: Discussed offline, this is a reasonable limitation

GeorgeNgMsft (approved):

PR #2821, opened 2026-08-07
Convert DroidCall multi-action cases

datduyng:
Converts DroidCall multi-action rows into translation benchmark cases with closed schemas, lineage hashes, and dependency-aware ordering.

```console
$ pnpm run jest-esm --runInBand --testPathPattern=translationBench.droidCallBenchmark.spec.js
Result: 1 suite passed, 1 test passed, 0 failed
$ node dist/translationBench/public_datasets/DroidCall/droidCallBenchmark.js
Result: droidcall-test-0, strict order, 2 actions
```


ts/packages/benchmarks/src/translationBench/public_datasets/DroidCall/droidCallBenchmark.ts
```diff
@@ -0,0 +1,219 @@
...
+
+// Convert one source row without requiring the downloader or code parser.
+export function toDroidCallBenchmarkCase(
+    row: DroidCallSourceRow,
+    split: "train" | "test",
+    rowIndex: number,
+): DroidCallBenchmarkCase | undefined {
+    if (row.answers.length < 2) return undefined;
+
+    const tools = row.tools.map(toDroidCallFunctionTool);
```
GeorgeNgMsft, line 158: There does not appear to be enforcement where every action has a corresponding tool here (which might be intentional). I would just make sure that logic exists somewhere, otherwise you may get unexpected failures or incorrectly reported data.

GeorgeNgMsft (approved):
Lgtm, minor comment

ts/packages/benchmarks/src/translationBench/public_datasets/DroidCall/droidCallBenchmark.ts
datduyng: Fixed in 83d3bf051. Rows are excluded when any gold action is absent from the candidate tools. Corpus audit: 2,656 valid multi-action rows, 26 excluded, zero conversion failures.

PR #2825, opened 2026-08-07
benchmarks: add DroidCall official scoring slice

datduyng:
- Add the DroidCall paper, released-code, and TypeAgent-adjusted scorer.
- Document the stdin contract and MIME presence-only adjustment.
- Keep the public-dataset slice at exactly 300 non-test lines.


ts/packages/benchmarks/src/translationBench/public_datasets/DroidCall/eval/officialDroidCallGrader.ts
```diff
@@ -0,0 +1,361 @@
...
+    semanticScorer: DroidCallSemanticScorer,
+): {
+    score: number;
+    correct: number;
+    total: number;
+    callSoftTotal: number;
+    callCount: number;
+} {
+    const responseByName = new Map(
+        row.response.map((response) => [response.name, response]),
```
GeorgeNgMsft: The DroidCall test set has multi-action rows that invoke the same API multiple times. That can be problematic here because map is only keyed by the API name so it will always take the result of your last call. So multi-call scenarios will check each call against the last result resulting in incorrect partial credit scores, extra credit, or unexpected failures.

ts/packages/benchmarks/src/translationBench/public_datasets/DroidCall/eval/officialDroidCallGrader.ts
```diff
@@ -0,0 +1,361 @@
...
+            throw new Error(`Unknown API '${answer.name}'`);
+        }
+        const call = scoreCall(
+            answer,
+            responseByName.get(answer.name),
+            api,
+            config,
+            semanticScorer,
+        );
+        failed ||= call.missing;
```
GeorgeNgMsft: DroidCall's released soft-accuracy metric doesn't remove credit for the row entirely and allows partial credit. The logic here seems to be all or nothing. We should evaluate whether this makes sense and ensure the methodology is correctly aligned.

GeorgeNgMsft (approved):
Please review the suggestions, otherwise looks good!

PR #2826, opened 2026-08-07
Support tracking cached tokens in aiclient

datduyng:
- Add optional `cached_tokens` on `CompletionUsageStats` (OpenAI-style: **subset** of `prompt_tokens`, not additive).
- Flatten Azure/OpenAI nested cache fields:
  - chat: `prompt_tokens_details.cached_tokens`
  - responses: `input_tokens_details.cached_tokens`
- Normalize Anthropic Messages usage to the same shape:
  - Anthropic reports exclusive buckets: `input_tokens + cache_creation_input_tokens + cache_read_input_tokens`
  - We map `prompt_tokens = sum of all three` and `cached_tokens = cache_read_input_tokens`

**Accounting convention (verified):**
- OpenAI: `input_tokens`/`prompt_tokens` already includes cache reads; `cached_tokens` is a subset (live LiteLLM probe: 2nd call `input_tokens=2649`, `cached_tokens=2304`).
- Anthropic: `input_tokens` is only the uncached tail; total input requires summing cache create/read buckets (docs). Adapters fold to OpenAI-style so benchmark cost/eval totals are comparable across providers.

We need this to compute benchmark cost per run and per eval case.


ts/packages/aiclient/src/apiTypes.ts
```diff
@@ -29,6 +29,8 @@ export type CompletionUsageStats = {
     prompt_tokens: number;
     // Total tokens (prompt + completion)
     total_tokens: number;
+    // Subset of prompt_tokens served from provider prompt cache (when reported).
```
TalZaccai: This says `cached_tokens` is a subset of `prompt_tokens`, but we already define this field twice with the opposite meaning:

- `dispatcher/types/src/dispatcher.ts:71` - "reported separately from prompt_tokens", and `total_tokens` is "prompt + completion, plus cached_tokens when reported"
- `agentSdk/src/action.ts:56` - same wording

Those three types are meant to be structurally compatible (see the note at `action.ts:47`), and `reasoning/reasoningLoopBase.ts:412` already implements the additive version (`total = input + output + cached`).

Please match the existing convention: cached tokens are their own bucket, not part of `prompt_tokens`. The rest of my comments assume that direction.

ts/packages/aiclient/src/wireApiProvider/shared.ts
```diff
@@ -74,6 +75,7 @@ export function usageFromInputOutput(
         prompt_tokens: prompt,
         completion_tokens: completion,
         total_tokens: prompt + completion,
+        ...(cachedTokens !== undefined && { cached_tokens: cachedTokens }),
```
TalZaccai, line 83: The new commit fixes the Anthropic numeric total by folding cache creation/read into `prompt_tokens` (2400 + 4 = 2404). That removes the old 904 undercount, but it does so by adopting the opposite contract from `dispatcher/types`, `agentSdk`, and `reasoningLoopBase.ts:412`.

To match the existing separate-bucket contract, this helper still needs `total_tokens: prompt + completion + (cachedTokens ?? 0)`. Otherwise normalizing Messages back to that shared contract will drop cached tokens from the total.

ts/packages/aiclient/src/wireApiProvider/messagesWireApiProvider.ts
```diff
@@ -58,6 +58,14 @@ function toTools(schemas: FunctionCallingJsonSchema[]): Tool[] {
     });
 }
 
+/** Anthropic cache hits are top-level cache_read_input_tokens (null when absent). */
+function cachedFromMessagesUsage(
```
TalZaccai: The new commit no longer drops `cache_creation_input_tokens` from the numeric total, but it folds both cache buckets into `prompt_tokens` and exposes only cache reads as `cached_tokens`. That still conflicts with the existing separate-bucket convention.

Anthropic's `input_tokens` already excludes both cache buckets, so keep it as `prompt_tokens`, set `cached_tokens` to read + creation, and let `shared.ts` add cached into the total.

This normalization also remains independent from `claude.ts:1788`, `claude.ts:2099`, `copilot.ts:1789`, and `copilot.ts:2145`, which use read + creation as the separate cached bucket. Extracting one shared helper would prevent this divergence.

ts/packages/aiclient/src/wireApiProvider/chatCompletionsWireApiProvider.ts
```diff
@@ -55,9 +64,19 @@ type ChatCompletionDelta = {
 type ChatCompletionChunk = {
     id: string;
     choices: ChatCompletionDelta[];
-    usage?: CompletionUsageStats;
+    usage?: ProviderUsage | null;
 };
 
+function flattenUsage(usage: ProviderUsage): CompletionUsageStats {
```
TalZaccai, line 70: OpenAI/Azure include cached tokens *inside* `prompt_tokens`, so once `cached_tokens` becomes its own bucket this needs `prompt_tokens: usage.prompt_tokens - cached`. `total_tokens` is unaffected (100 -> 60 + 20 + 40 = 120), so it still matches the provider's reported total - only the split changes.

Separately: `flattenUsage` and `shared.ts:usageFromInputOutput` are now two builders of the same type with different `total_tokens` rules, so the fix above has to land in both. Worth collapsing into one helper.

ts/packages/aiclient/src/wireApiProvider/responsesWireApiProvider.ts
```diff
@@ -88,7 +95,11 @@ export class ResponsesWireApiProvider implements ProviderAdapter {
 
     extractUsage(data: unknown): CompletionUsageStats | undefined {
         const usage = (data as ResponsesResult).usage;
-        return usageFromInputOutput(usage?.input_tokens, usage?.output_tokens);
+        return usageFromInputOutput(
+            usage?.input_tokens,
+            usage?.output_tokens,
+            usage?.input_tokens_details?.cached_tokens,
```
TalZaccai, line 102: Same as chat completions: `input_tokens_details.cached_tokens` is already counted inside `input_tokens`, so this needs to subtract before passing it through, otherwise cached tokens get counted twice once `shared.ts` adds them into the total.

ts/packages/aiclient/src/wireApiProvider/responsesWireApiProvider.ts
```diff
@@ -116,6 +124,7 @@ export class ResponsesWireApiProvider implements ProviderAdapter {
                     piece.usage = usageFromInputOutput(
                         usage.input_tokens,
                         usage.output_tokens,
+                        usage.input_tokens_details?.cached_tokens,
```
TalZaccai, line 128: Same expression as line 101. When the subtraction goes in, it needs to happen here too - easy to fix one and miss the other. Consider a small `cachedFromResponsesUsage(usage)` so both paths share it.

ts/packages/aiclient/test/providerAdapter.spec.ts
```diff
@@ -185,6 +185,26 @@ describe("ChatCompletionsWireApiProvider (default path is byte-identical)", () =
         );
     });
 
+    test("extractUsage tracks nested prompt_tokens_details.cached_tokens", () => {
```
TalZaccai, line 188: With cached tokens split out of `prompt_tokens`, this becomes `prompt_tokens: 60, completion_tokens: 20, cached_tokens: 40, total_tokens: 120`.

The existing "returns the usage block verbatim" test already covers `prompt_tokens_details` missing entirely. The remaining edge case is `prompt_tokens_details` present without `cached_tokens`; that should also omit the field.

ts/packages/aiclient/test/providerAdapter.spec.ts
```diff
@@ -314,6 +334,43 @@ describe("MessagesWireApiProvider", () => {
         });
     });
 
+    test("extractUsage tracks cache_read_input_tokens as cached_tokens", () => {
```
TalZaccai: This is the case that surfaces the bug: expected should be `prompt_tokens: 900, completion_tokens: 4, cached_tokens: 1500, total_tokens: 2404` - 1500 being read (700) + creation (800), and the total including them.

ts/packages/aiclient/test/providerAdapter.spec.ts
```diff
@@ -382,6 +439,23 @@ describe("ResponsesWireApiProvider", () => {
         });
     });
 
+    test("extractUsage tracks nested input_tokens_details.cached_tokens", () => {
```
TalZaccai, line 463: Becomes `prompt_tokens: 86, completion_tokens: 300, cached_tokens: 1920, total_tokens: 2306` - total unchanged, just split correctly.

Across all three adapters it'd be worth asserting `prompt_tokens + completion_tokens + (cached_tokens ?? 0) === total_tokens`. That invariant is the thing that makes the field usable for cost math, and it's what currently fails on the messages path.

ts/packages/aiclient/src/wireApiProvider/shared.ts
```diff
@@ -64,6 +64,7 @@ export type { Filter, FilterError, FilterResult };
 export function usageFromInputOutput(
     inputTokens: number | undefined,
     outputTokens: number | undefined,
+    cachedTokens?: number,
```
TalZaccai, line 72: Heads up that the field won't reach a benchmark yet: roughly 13 places hand-sum usage three fields at a time, and only `dispatcher/src/execute/actionHandlers.ts:405` knows about `cached_tokens`. Notably `aiclient/src/tokenCounter.ts:24` doesn't, and neither does `interpretRequest.ts:256`.

A shared `addUsage(acc, usage)` in aiclient would stop the next field from repeating the same sweep.

TalZaccai (changes requested):

ts/packages/aiclient/src/wireApiProvider/chatCompletionsWireApiProvider.ts
GeorgeNgMsft: We were just discussing this case offline, good catch! Beat me to the punch haha

ts/packages/aiclient/src/wireApiProvider/shared.ts
datduyng: Resolved

ts/packages/aiclient/src/wireApiProvider/messagesWireApiProvider.ts
datduyng: resolved

ts/packages/aiclient/src/wireApiProvider/responsesWireApiProvider.ts
datduyng: Resolved

ts/packages/aiclient/src/wireApiProvider/shared.ts
datduyng: Yes, this is to get the count to the synthesizer only for now which is part 2/n and 3n. Thanks!

TalZaccai (approved):
LGTM

TalZaccai (approved):

PR #2830, opened 2026-08-08
[OTEL] Create TypeAgent Span Structure

GeorgeNgMsft:
Defines the stable OpenTelemetry contract for TypeAgent’s core request trace:

typeagent.request
├── typeagent.translation
├── typeagent.reasoning
├── typeagent.action
└── typeagent.llm

Adds allowlisted, redacted attributes for agent, action, model, session, activation, and legacy trace correlation. Also adds an in-memory test harness for asserting emitted spans and parent-child relationships. 

Depends on : https://github.com/microsoft/TypeAgent/pull/2814


datduyng (approved):

datduyng:
Approve. I don't think we should use the term "harness". I have worked with otel for a while. haven't heard that term. harness is very new like around April where agent was getting popular

PR #2836, opened 2026-08-10
Preserve chat state

jebrans:
## Problem
 
 The `mobile-2` Android sample kept its connection and transcript in `MainActivity`. Activity recreation—such as rotation, theme, font-scale, or locale changes—therefore disconnected the WebSocket and discarded the visible chat. The transcript was also lost whenever Android reclaimed the app process.
 
 The Android client does not currently load AgentServer display history, so it needs a local conversation cache to restore the UI immediately after process death.
 
 ## Changes
 
 - Move connection and chat ownership into `ChatViewModel` so configuration changes preserve the WebSocket, messages, input state, and pending interactions.
 - Add `ConversationStore`, backed by `SharedPreferences`, to persist:
   - the visible transcript;
   - the last joined AgentServer conversation ID.
 - Resume the saved conversation directly by passing its ID to `joinConversation` during reconnect.
 - If the saved conversation no longer exists, discard its stale local transcript and fall back to the server’s default conversation.
 - Align Android naming with AgentServer terminology: **conversation** represents user-facing identity and chat history, while **session** is reserved for dispatcher runtime state.
 - Replace the misleading **New chat** action with **Clear chat**. This clears only the local display history; the server-side conversation and agent memory are unchanged.
 - Seal restored messages as final so they cannot remain stuck in a streaming state or be modified by later display updates.
 - Debounce persistence writes during streaming and synchronously flush the final state when the ViewModel is cleared.
 - Limit persisted data to the newest 200 messages and remove messages older than 30 days on both save and load.
 - Exclude the persisted transcript from Android Auto Backup and device-to-device transfer.
 - Buffer agent-driven Android actions across Activity recreation and briefly wait for the recreated Activity to reach `RESUMED` before dispatching foreground intents.
 - Automatically send final voice-recognition results.
 
 ## Testing
 
 - Added JVM tests for conversation serialization, retention, message limits, legacy payloads, clock changes, and missing-conversation error detection.
 - Verified transcript restoration after configuration changes, background process reclamation, and force-stop.
 - Verified expired messages are removed from both the UI and persisted storage.
 - Verified **Clear chat** removes the local transcript without changing the AgentServer conversation.
 - Verified end-to-end behavior against a live AgentServer through a Dev Tunnel.
 
 ## Follow-up
 
 A later PR can treat `ConversationStore` as an immediate/offline cache and use `getDisplayHistory(afterSeq?)` to backfill updates produced by other clients sharing the same conversation.


android/samples/mobile-2/app/src/main/java/com/example/typeagentchat/WebSocketManager.kt
GeorgeNgMsft: I see you've added joinedConversationId, since it persists after a disconnect, maybe it should be renamed as lastJoinedConversationId or recentConversationId.

I think you can just pass the id back into the connection as part of the connection options for a natural reconnect, and if it doesn't exist, then you can just handle the fallback logic. If the conversation no longer exist, it should create a new conversation as the fallback and clear then. 

That way you don't need to unnecessarily load history and reconcile it -- matching the existing AgentServer conventions. 

This may also be helpful reference https://github.com/microsoft/TypeAgent/blob/main/ts/docs/architecture/agents/agentServerConversations.md

android/samples/mobile-2/app/src/main/java/com/example/typeagentchat/ConversationStore.kt
```diff
@@ -0,0 +1,257 @@
+package com.example.typeagentchat
```
GeorgeNgMsft, line 1: On AgentServer, the chats are stored as [ConversationRecord objects ](https://github.com/microsoft/TypeAgent/blob/89afe71316c771ac8c60dd1ca109ad22cba4ad57/ts/packages/agentServer/server/src/conversationManager.ts#L131) and we persist them on disk in a conversations.json file. We should try to preserve the 'conversation' terminology here so that we match AgentServer. 

The reason is that 'sessions' have been reserved for Dispatcher's runtime state which also gets persisted onto disk with configurations, cache setups, etc. This is a legacy functionality that I don't personally think is justified anymore (and it adds so much confusion), but cleaning it up should be a separate discussion. 

Conversation : User-facing identity & chat history
Session : DIspatcher configurations, caches, agent states

So we should probably rename the classes, files, etc. since these involve the user conversations rather than dispatcher sessions.

android/samples/mobile-2/app/src/main/java/com/example/typeagentchat/ChatViewModel.kt
```diff
@@ -0,0 +1,304 @@
...
+
+    /**
+     * Starts a new chat: clears the on-screen transcript and the copy on disk.
+     *
+     * This is a client-side reset only. `joinConversation` keeps returning the
+     * same server-side conversation, so the agent's own memory is untouched -
+     * the server exposes no RPC to start a fresh one.
+     */
+    fun startNewChat() {
+        webSocketManager.clearMessages()
```
GeorgeNgMsft, line 342: Let's rename this to "clearChatHistory" or "clearChat", I think start newChat is a bit misleading here. 

On the other TypeAgent canvases, @clear usually has some client-side implementation for clearing the display history without touching the memory or conversation (UX only). Reconnecting will reload the history though (not sure if clearing the on disk memory is a good idea). 

New chat will typically follow this flow : 
createConversation
→ joinConversation(new ID)
→ rebind and persist new ID
→ leaveConversation(old ID)
→ clear/replay UI

CLI Implementation 
https://github.com/microsoft/TypeAgent/blob/05562958a9261c3b44c1f6e0dfcb1141625d30f4/ts/packages/cli/src/conversationCommands.ts#L317

AgentServer client reference https://github.com/microsoft/TypeAgent/blob/05562958a9261c3b44c1f6e0dfcb1141625d30f4/ts/packages/agentServer/client/src/conversation/lifecycle.ts#L190

android/samples/mobile-2/app/src/main/java/com/example/typeagentchat/ConversationStore.kt
GeorgeNgMsft: **[For a later PR]** Currently, it seems like ChatSessionStore as the authoritative storage. However, a big capability we want to continue supporting is the ability to have real time-shared conversations across clients. Right now, multiple clients are able to connect to a conversation and load the full conversation history by calling the **getDisplayHistory** rpc then get updates broadcasted to them.

**Currently Here:**

```
Load local messages
→ join an implicit default conversation
→ compare IDs
→ receive only future live events
→ overwrite local snapshot
```

This currently supports live display events from other clients, but there is potential for a display history gap since the Android client doesn't read the display history. I propose that we treat the ChatSessionStore as a cache for immediate/offline display and display log cursor.

**Proposed :** 

```
Load local cache + saved conversation ID + sequence cursor
→ join that exact conversation
→ fetch missing server display-history entries via getDisplayHistory(afterSeq?)
→ reconcile/rebuild the cache
→ receive future live events
```

This will let you show the display history immediately while also backfilling any chat history from other clients on the same conversations. 

---

This can be done later on and discussed further if you'd like. The current flow is fine for this PR

GeorgeNgMsft (changes requested):
Just to restate the current layering that I see : 
**ChatViewModel** coordinates lifecycle + UX, which should look similar to the CLI/Shell reference flows
**WebSocketManager** is the Android agent-server RPC client

It would be worth matching the persistence semantics in AgentServer's existing canvases now (the additional functionality can be built out later on)

I think this should include : 
- "New conversation/chat" flow alignment with existing flows
- "Conversation" naming alignment for better semantic representation
- Updating the conversation resume behavior (for efficiency)

Overall, I think this is great! The conversation/session naming gets a bit confusing so I just wanted to help draw the distinctions early.

android/samples/mobile-2/app/src/main/java/com/example/typeagentchat/WebSocketManager.kt
jebrans: Renamed to lastJoinedConversationId. The saved id is now passed into connect(resumeConversationId = …) and forwarded as joinConversation's conversationId connect option, so the client rejoins the exact conversation the transcript belongs to. When it's null the option is omitted entirely, so a first run still auto-joins the default.

android/samples/mobile-2/app/src/main/java/com/example/typeagentchat/ChatViewModel.kt
jebrans: Renamed to clearChatHistory() This will clear local cache and not Typeagent's data.

android/samples/mobile-2/app/src/main/java/com/example/typeagentchat/ConversationStore.kt
jebrans: Renamed. Will add a follow up for ChatSessionStore

GeorgeNgMsft (approved):
Looks good! Thanks for resolving the comments

GeorgeNgMsft:
Something worth noting, for the other clients, we fall back to a "named default" of "CLI" for CLI and "Shell" for Shell. It's a completely arbitrary behavior which is a bit different from the "default" being used here. It's worth noting just because it differs in behavior which may come into play when Android joins the multi-client scenario fully. It is probably a good idea to create an "Android" default conversation at this point so the behavior is unified.

At some point we will probably change it though.

PR #2837, opened 2026-08-10
[OTEL] Wire root request span

GeorgeNgMsft:
Adds the dispatcher’s root OpenTelemetry request span, establishing the parent for future translation, reasoning, action, and LLM spans.

• Creates one  typeagent.request  span per command.
• Adds session, activation, and legacy trace correlation attributes.
• Starts a new trace by default.
• Supports opt-in host trace parenting with  telemetry.joinActiveTrace .
• Preserves the submitting host’s context across queued requests.
• Records privacy-safe exception classifications by default.
• Supports explicit redacted diagnostic capture with  telemetry.captureSensitiveErrorDetails .
• Marks cancellations distinctly using  AbortError  and the  "cancelled"  status message.
• Preserves existing cancellation, cleanup, and command-result behavior.
• Adds coverage for span lifecycle, parenting, queue propagation, cancellation, errors, and privacy defaults.

When running Grafana locally, we can see the base trace of our dispatcher requests now via Tempo
image


ts/packages/dispatcher/dispatcher/src/otel/rootRequestSpan.ts
```diff
@@ -0,0 +1,122 @@
...
+export interface RootRequestSpanOptions {
+    /**
+     * Explicit parent context. Defaults to ROOT_CONTEXT so unrelated ambient
+     * instrumentation cannot adopt the TypeAgent request span accidentally.
+     */
+    readonly parentContext?: Context | undefined;
+    readonly captureSensitiveErrorDetails?: boolean | undefined;
+}
+
+/**
```
datduyng: In general, we should go light with documentation like this. since there are no way to validate it, doc comment drift from code which often take exdtra turns for agents.

We should remove this large comment block

ts/packages/dispatcher/dispatcher/src/command/command.ts
```diff
@@ -356,6 +364,23 @@ export async function processCommandNoLock(
...
+        // converts an exception to `ActionResult`, it must still record
+        // the exception and error status." The catch below converts the
+        // thrown error into a user-visible display message + logged event,
+        // not a re-throw; without this recordException/setStatus pair the
+        // root span would silently end with status UNSET on real failures.
+        const activeSpan = trace.getActiveSpan();
+        if (activeSpan !== undefined) {
+            otel.recordTypeAgentSpanException(activeSpan, e, {
+                safeName: "CommandError",
+                safeMessage: "command failed",
```
datduyng: fundamentally, sensitive options and handling should be done at source. Each consumer of recordTypeAgentSpanException won't know how to use recordTypeAgentSpanException and stick to to easiest option which is don't pass anything. 

I wouldn't expose this to consumer

ts/packages/dispatcher/dispatcher/src/command/command.ts
```diff
@@ -453,38 +479,101 @@ export async function processCommand(
...
-                abortController.signal,
-            );
-            context.clientIO.setUserRequest(requestId, originalInput);
+    // Compute the correlation attributes ONCE, before opening the root
+    // `typeagent.request` span. Values that only become known later
+    // (agent name, action name) are set on the active span by downstream
+    // steps in later phases; the root span carries only the values known
+    // at the outermost async boundary. Everything the wrapper receives is
+    // an identifier, not user text - see setTypeAgentSpanAttributes.
+    const sessionId = context.session.sessionDirPath
```
datduyng, line 483: how can session ever be undefined? if it’s undefined it's a bug. I feel like we should always have a session.

datduyng (approved):

ts/packages/dispatcher/dispatcher/src/command/command.ts
GeorgeNgMsft: This is because the sessionId is retrieved from the directory backing the persistent data. Currently ephemeral sessions (that aren't persisted to disk) don't have an ID. Session itself has no ID field; it's something we can change later but not here.

ts/packages/dispatcher/dispatcher/src/command/command.ts
GeorgeNgMsft: Yeah, I went back and forth a few times on this, but I agree, let's do stick with that.

PR #2849, opened 2026-08-12
Update PowerShell dev actions and namespace handlers

hillary-mutisya:
- add cancellation, structured failures, bounded repair, and flow reuse
- register and implement all PowerShell namespaces
- require confirmation for mutating actions with unattended default-deny
- extract namespace actions into typed handlers with a shared registry
- add routing, persistence, concurrency, sandbox, and runtime coverage
- update Copilot dev-action routing and project documentation


ts/packages/agents/powershell/scripts/scriptHost.ps1
```diff
@@ -73,26 +113,28 @@ try {
...
-                                break
-                            }
-                        }
-                        # ENFORCEMENT: Block execution if path not allowed
-                        if (-not $pathAllowed) {
-                            Write-Error "Path access denied: '$resolvedPath' is not in allowedPaths. Allowed paths: $($expandedAllowedPaths -join ', ')"
-                            exit 1
-                        }
+                try {
+                    $resolvedPath = Get-CanonicalFileSystemPath $val
```
jebrans: Will this work for URL's in data like 
```
{
    actionName: "writeFile",
    parameters: {
        path: "C:\\Users\\me\\notes.txt",
        content: "https://example.test/api",
    },
}
```

ts/packages/agents/powershell/scripts/scriptHost.ps1
```diff
@@ -73,26 +113,28 @@ try {
                     continue
```
GeorgeNgMsft: Copilot highlighted a concern here that Start-Process accepts executable names and resolves them through PATH. So executables outside of allowedPaths can still be run here for example : 

` Path = "powershell.exe"  with  Arguments = "-NoProfile -Command Get-Process"`

It suggests explicitly declaring which action parameters represent paths. This will allow executable parameters to be resolved to canonical paths and validated, while parameters representing things like library names can properly skipped and not marked as a path.

ts/packages/agents/powershell/src/actionHandler.mts
```diff
@@ -477,36 +408,272 @@ async function executeDraftRecipe(
...
+        };
+        const execution = await executeDraftRecipe(
+            candidate,
+            executionParameters.parameters,
+            context.abortSignal,
+        );
+        if ("error" in execution) {
+            return execution.error;
+        }
+        context.abortSignal?.throwIfAborted();
```
GeorgeNgMsft: We may want to split up the post-execution work of cancellation check, updateFlowScript(), and recordUsage() and separate their potential failure scenarios. 

For example, update failures (after executing) or cancellation triggered _after execution (where it may have already changed state)_ should perhaps return a partialFailure to surface potential side effects (but no rollback).  This should probably be separate from script failure scenarios where you do want a rollback.

ts/packages/agents/powershell/src/actionHandler.mts
```diff
@@ -731,6 +841,19 @@ async function handlePowerShellFlowAction(
...
+                );
+            }
+            return repairAndExecutePowerShellFlow(
+                action.parameters as Record<string, unknown>,
+                flowStore,
+                context,
+            );
+        }
+
         case "editPowerShellFlow": {
```
GeorgeNgMsft, line 1025: So in the earlier repair operation it will acquire lock using **withFlowMutationLock(flowName)**, but editPowerShellFlow here does not acquire the lock. So, it is hypothetically possible (but unlikely) that repair + edit can race to mutate. It would probably be a good idea to have edit also acquire lock too.

GeorgeNgMsft (changes requested):
Left a few comments, I think two are optional, but the potential sandbox gap is worth addressing.

ts/packages/agents/powershell/scripts/scriptHost.ps1
hillary-mutisya: Ah, yes. The code would inadvertently capture the URL because we relied on string shape. I've updated the code now to remove string-shape path inference and validate only parameters explicitly declared as paths.

ts/packages/agents/powershell/scripts/scriptHost.ps1
hillary-mutisya: Good catch. The fix was to add an executable parameter role, resolve bare executable names through PowerShell command resolution, canonicalize the resolved file, and enforce `allowedPaths`.

ts/packages/agents/powershell/src/actionHandler.mts
hillary-mutisya: Fixed

ts/packages/agents/powershell/src/actionHandler.mts
hillary-mutisya: Fixed

GeorgeNgMsft (approved):

PR #2855, opened 2026-08-12
Add cross-process TPM rate limiter

datduyng:
Shared SQLite ledger so concurrent translation-bench processes stay under a rolling-window token budget.

### DIFF
--- a/ts/packages/benchmarks/src/core/rateLimiter.ts
+++ b/ts/packages/benchmarks/src/core/rateLimiter.ts
@@ -1,0 +1,318 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import fs from "node:fs";
+import path from "node:path";
+import { randomUUID } from "node:crypto";
+import { DatabaseSync, type StatementSync } from "node:sqlite";
+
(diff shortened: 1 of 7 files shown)


GeorgeNgMsft (approved):

ts/packages/benchmarks/src/core/rateLimiter.ts
```diff
@@ -0,0 +1,318 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import fs from "node:fs";
+import path from "node:path";
+import { randomUUID } from "node:crypto";
+import { DatabaseSync, type StatementSync } from "node:sqlite";
```
GeorgeNgMsft, line 7: Copilot noted that we use better-sqlite3 elsewhere in the codebase, but this introduces a second implementation of sqlite with a different package. Please consider unifying to the existing standard

datduyng: it's better to use built in sqlite when it possible. Discussed offline. I will keep this as simple as possible to use node built in.

PR #2858, opened 2026-08-12
Support Python literals in public datasets

datduyng:
Supports bounded parsing of the Python literal subset used by public tool datasets, including safe numeric lexeme preservation and prototype-safe dictionaries.

```console
Input:  {'enabled': True, 'items': [None, 1.230e+04]}
Output: { enabled: true, items: [null, 12300] }
Test:   PASS translationBench.pythonLiteral.spec.js (3 passed, 0 failed)
```


ts/packages/benchmarks/src/translationBench/public_datasets/pythonLiteral.ts
```diff
@@ -0,0 +1,288 @@
...
+            f: "\f",
+            n: "\n",
+            r: "\r",
+            t: "\t",
+            v: "\v",
+        };
+        if (character in simpleEscapes) {
+            return { value: simpleEscapes[character]!, end: offset + 1 };
+        }
+        const width = character === "x" ? 2 : character === "u" ? 4 : 0;
```
GeorgeNgMsft: I think Python supports eight digit unicode escapes for like '\U0001f600' and this is being missed right now. Should double check that and cover it as well if it is

ts/packages/benchmarks/src/translationBench/public_datasets/pythonLiteral.ts
```diff
@@ -0,0 +1,288 @@
...
+        const match = NUMBER_PATTERN.exec(this.text);
+        if (match === null) {
+            throw this.error("unexpected Python literal token", offset);
+        }
+        const lexeme = match[0];
+        const number = Number(lexeme);
+        if (
+            !this.preserveNumberLexemes &&
+            (!Number.isFinite(number) ||
+                (Number.isInteger(number) && !Number.isSafeInteger(number)))
```
GeorgeNgMsft, line 272: Some scientific numbers like 1e10 or 101.0 can pass the first Number.isInteger(number) check in python, but Number.isSafeInteger will reject it. Might only want to conditionally enforce the isSafeInteger() check on numbers without decimals or "e". Small edge case though

GeorgeNgMsft (approved):
Left a couple comments on niche edge cases to consider

ts/packages/benchmarks/src/translationBench/public_datasets/pythonLiteral.ts
datduyng: Fixed in 8865e3e9663c2d257d4522cd2fcf4dd106174072

ts/packages/benchmarks/src/translationBench/public_datasets/pythonLiteral.ts
datduyng: Fixed in 8865e3e9663c2d257d4522cd2fcf4dd106174072

PR #2859, opened 2026-08-12
Add translation-bench gold-action validation helper

datduyng:
Validate gold actions after restoring required empty parameters and single-literal union fields.

### DIFF
--- a/ts/packages/benchmarks/src/translationBench/synthesizer/actionValidation.ts
+++ b/ts/packages/benchmarks/src/translationBench/synthesizer/actionValidation.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 import {
+    resolveTypeReference,
     validateAction,
     type ActionSchemaTypeDefinition,
 } from "@typeagent/action-schema";
@@ -25,9 +26,9 @@ export function translationBenchActionValidationPayload(
     for (const [name, field] of Object.entries(definition.type.fields)) {
         if (name === "actionName" || name === "parameters") continue;
         if (field.optional) continue;
-        const fieldType = field.type;
+        const fieldType = resolveTypeReference(field.type);
         if (
-            fieldType.type === "string-union" &&
+            fieldType?.type === "string-union" &&
             fieldType.typeEnum.length === 1
         ) {
             payload[name] = fieldType.typeEnum[0];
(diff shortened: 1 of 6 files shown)


ts/packages/benchmarks/src/translationBench/synthesizer/actionValidation.ts
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) Microsoft Corporation.
```
TalZaccai, line 1: Helper isn't called anywhere, so the bug is still live — `generationCandidate.ts:233` still drops `parameters` when undefined, `benchmark.ts:2279` passes the raw action. Can we wire those up here? Also needs a spec next to `goldParameterHygiene`.

ts/packages/benchmarks/src/translationBench/synthesizer/actionValidation.ts
```diff
@@ -0,0 +1,56 @@
...
+): Record<string, unknown> {
+    const payload: Record<string, unknown> = {
+        actionName: action.actionName,
+    };
+    for (const [name, field] of Object.entries(definition.type.fields)) {
+        if (name === "actionName" || name === "parameters") continue;
+        if (field.optional) continue;
+        const fieldType = field.type;
+        if (
+            fieldType.type === "string-union" &&
```
TalZaccai: Only catches inline unions. A `type-reference` to a single-literal alias gets skipped. Run it through `resolveTypeReference` (action-schema `utils.ts:35`) first.

TalZaccai (changes requested):

ts/packages/benchmarks/src/translationBench/synthesizer/benchmark.ts
```diff
@@ -2276,7 +2276,12 @@ function validateExpectedActions(
                 `${label} expects unknown existing TypeAgent action '${action.schemaName}.${action.actionName}'`,
             );
         }
-        validateAction(definition, action);
+        validateTranslationBenchGoldAction(definition, {
```
TalZaccai, line 2279: This used to pass the whole `action`, so `validateAction` caught stray top-level keys (`schemaName` was whitelisted). Now we rebuild the payload from just `actionName`/`parameters`, so a typo'd key in gold JSON gets silently dropped instead of erroring. Intentional? If not, the helper could take the full action and keep the extraneous check.

TalZaccai (approved):

PR #2861, opened 2026-08-12
Gate empty-gold negatives on utterance shape

datduyng:
Reject empty-gold utterances that are questions, sibling commands, or refuse-then-alternate.

This is a cheap and deterministic way to filter out utterances catching 90% of LLm generated utterances

### DIFF
--- a/ts/packages/benchmarks/src/translationBench/synthesizer/emptyGoldUtterance.ts
+++ b/ts/packages/benchmarks/src/translationBench/synthesizer/emptyGoldUtterance.ts
@@ -1,0 +1,128 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/** ; | em-dash | en-dash | spaced hyphen — never bare `.` (schema.action / domains). */
+const CLAUSE_SEP = String.raw`(?:[;]|\u2014|\u2013|\s-\s)`;
+
+/**
+ * Clause separators for multi-part empties. Deliberately excludes `.` so
+ * schema.action tags, domains, and abbreviations do not false-split.
+ */
+const CLAUSE_SPLIT_RE = new RegExp(String.raw`\s*${CLAUSE_SEP}\s*`);
+
+/**
+ * Trailing clauses that still mean abstain (not a new tool request).
+ * Stripped before secondary-clause checks.
+ */
+const ABSTAIN_TRAIL_RE = new RegExp(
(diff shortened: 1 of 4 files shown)


ts/packages/benchmarks/src/translationBench/synthesizer/emptyGoldUtterance.ts
```diff
@@ -0,0 +1,128 @@
...
+ * schema.action tags, domains, and abbreviations do not false-split.
+ */
+const CLAUSE_SPLIT_RE = new RegExp(String.raw`\s*${CLAUSE_SEP}\s*`);
+
+/**
+ * Trailing clauses that still mean abstain (not a new tool request).
+ * Stripped before secondary-clause checks.
+ */
+const ABSTAIN_TRAIL_RE = new RegExp(
+    String.raw`${CLAUSE_SEP}\s*(?:I\b[\s\S]*|let\s+it\b[\s\S]*|leave\b[\s\S]*?\b(?:alone|unchanged|untouched)\b[\s\S]*|keep\b[\s\S]*|stay\b[\s\S]*|so\b[\s\S]*|because\b[\s\S]*|since\b[\s\S]*)$`,
```
GeorgeNgMsft, line 18: Copilot is calling this out as some commands are being stripped unexpectedly even though they are valid : 

For example `Don't close this tab; keep scrolling down` will remove `keep scrolling down` and so the classification is incorrect as it views it as pure refusal `Don't close this tab` (when it's not).

datduyng: it's fine - we aren't trying to be hyper exact. this is data cleaning - we aren't trying to make everything perfect. typeagent right now have 58% FPR thus any improvement from this is good

GeorgeNgMsft (approved):
Approving, there is an improvement that can be made but as discussed offline, you are incrementally working on the data cleaning quality.

ts/packages/benchmarks/src/translationBench/synthesizer/emptyGoldUtterance.ts
datduyng: Another problem I faced

 that our LLM token rate limit is too low. Each experimentation loop I do take like 3 hours. so I need to swap out many of the filtering steps with a good enough filter

PR #2870, opened 2026-08-13
Complete MCP v2 extension management and security

hillary-mutisya:
Complete the MCP v2 integration by making MCP servers fully managed, secure, dynamic TypeAgent extensions.

- unify installed-agent and MCP runtime sources across all hosts
- persist normalized MCP configs with stable identity, provenance, trust, enabled state, tool policy, and credential references
- extend `@package` install, list, update, and uninstall with MCP support, previews, confirmation, ambiguity handling, and management commands
- add local MCP config imports and an MCP Registry v0.1 source with pagination, incremental caching, exact version resolution, remote HTTP, and verified npm package materialization
- safely track and remove owned package paths without deleting shared or unrelated files
- add credential-store and OAuth/PKCE integration points without persisting plaintext secrets
- enforce transport, command, domain, registry, package, and per-tool policies before installation, connection, and invocation
- require confirmation for potentially mutating MCP tools and replace Copilot's unconditional permission approval
- add sanitized, bounded audit logging for MCP lifecycle and tool activity
- validate MCP arguments and structured results using bounded JSON Schema 2020-12 validation with external references disabled
- preserve tool metadata under stable server-config/tool identities
- subscribe to tools/list_changed and atomically refresh schemas across all connected sessions through the dispatcher reload path
- add extensive coverage for installation transactions, registry behavior, security, OAuth, lifecycle fan-out, validation, and catalog refresh

Durable secret persistence and browser-based OAuth interaction remain host-injected interfaces. Default implementations use environment/session credentials and fail explicitly when durable secure storage is unavailable.

### DIFF
--- a/ts/packages/defaultAgentProvider/src/mcp/mcpAppAgentSource.ts
+++ b/ts/packages/defaultAgentProvider/src/mcp/mcpAppAgentSource.ts
@@ -12,10 +12,21 @@ import { createLimiter } from "@typeagent/common-utils";
 import {
     createMcpServerAppAgentProvider,
     McpClientInfo,
+    McpHostServices,
 } from "./mcpServerProvider.js";
-import { NormalizedMcpServerConfig } from "./mcpServerConfig.js";
+import {
+    NormalizedMcpServerConfig,
+    resolveTransportConfig,
+} from "./mcpServerConfig.js";
 import { McpServerStore } from "./mcpServerStore.js";
+import { McpConnection } from "./mcpConnection.js";
 import registerDebug from "debug";
+import { enforceMcpPolicy } from "./mcpPolicy.js";
+import { defaultMcpPolicy } from "./mcpPolicy.js";
+import type { McpPolicy } from "./mcpPolicy.js";
+import { McpOAuthProvider, getMcpAuthState } from "./mcpOAuth.js";
+import { SessionMcpCredentialStore } from "./mcpCredentialStore.js";
+import { nullMcpAuditSink } from "./mcpAudit.js";
 
 const debug = registerDebug("typeagent:mcp:source");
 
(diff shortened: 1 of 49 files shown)


ts/packages/defaultAgentProvider/src/mcp/mcpServerProvider.ts
```diff
@@ -1,139 +1,694 @@
...
+        config.toolApproval?.persisted?.[toolName] ??
+        (config.toolApproval?.deny?.includes(toolName)
+            ? "deny"
+            : config.toolApproval?.allow?.includes(toolName)
+              ? "allow"
+              : undefined)
+    );
+}
+
+function shouldPromptForTool(
```
GeorgeNgMsft, line 179: The decision for when a tool should be prompting should probably be left to the User's explicit approval or the calling host. We probably shouldn't decide solely based on the MCP Server's self-reporting because : 

1. The server we're calling could hypothetically be malicious and falsely report its safety
2. The MCP server could just be incorrectly labeled leading to unexpected destructive behavior

It's probably better to use the annotations to be able to only increase the restrictiveness of the gate but not reduce it. Shifting the standard gating decision to the host/user will help improve our security stance.

ts/packages/defaultAgentProvider/src/mcp/mcpServerProvider.ts
```diff
@@ -1,139 +1,694 @@
...
+    configured: "allow" | undefined,
+    sessionApprovals: Map<string, Set<string>>,
+    parameters: Record<string, unknown> | undefined,
+    sensitiveValues: string[],
+): Promise<McpToolDecision> {
+    const sessionId = context.sessionContext.sessionContextId;
+    if (
+        configured === "allow" ||
+        !shouldPromptForTool(config, tool) ||
+        sessionApprovals.get(sessionId)?.has(toolId)
```
GeorgeNgMsft, line 204: This is kind of a low priority/unlikely scenario in the life of a single session but something to consider : 

I see that we check the sessionApprovals for the existence of the toolId in the session. However, there is a potential scenario where a MCP server could update their tool and the behavior changes, but it's still approved. Ideally, it'll need to reacquire approval, but there isn't a great way of doing that right now. 

I do see below that we check for "toolsChanged", in terms of updates to the catalog. It may be worth considering if we can track something similar for tool versioning to refresh approval requirements, but it's more of a nice todo in the future.

ts/packages/defaultAgentProvider/src/mcp/mcpAppAgentSource.ts
```diff
@@ -96,78 +146,283 @@ export function createMcpAppAgentSource(
...
-                await fanOut(clients, issuingController, async (mutation) => {
-                    // Replace in place: remove the old provider before adding the
-                    // new one so the name is never registered twice.
-                    if (existing !== undefined) {
-                        await mutation.removeProvider(existing, {
-                            dropConfig: false,
-                        });
-                    }
-                    await mutation.addProvider(next);
+                await replaceUserConfig(config, issuingController);
```
GeorgeNgMsft, line 249: A potential edgecase here is that replaceUserConfig can commit the config successfully, but the following services.audit.write() can still fail for artbirary reasons (like running out of space, permissions, locks, etc). 

If the write() fails, it'll throw an exception in the parent tryCatch during runMcpInstall, triggering cleanup() which deletes the MCP package files. 

This can lead to an awkward situation where the MCP server appears in the list, but it won't actually function because the package files have been cleaned up.

You may want to make the audit best effort but not trigger the cleanup rollback. Alternatively, the rollback behavior should probably remove the config as well. Either option seem reasonable but would be good to clean up.

GeorgeNgMsft (approved):
Overall looks good, left a couple comments for edge cases we should at least note or address (whether in this PR or later).

PR #2877, opened 2026-08-14
Add explicit macro trace capture

hillary-mutisya:
Adds the first macro milestone: explicit, one-interaction trace capture.

- Adds `@typeagent macro record`, `cancel`, and `status`.
- Claims the next interaction without changing the active TypeAgent mode.
- Normalizes transcript tool calls, results, failures, and permission denials.
- Redacts sensitive values before persistence.
- Stores traces atomically under the agent-server instance directory.
- Adds typed agent-server recording RPCs that do not require a conversation.
- Rejects incomplete, mismatched, expired, or duplicate captures.

This PR creates immutable traces only. Macro induction, catalog management, and execution remain out of scope.

## Validation

- Macro core tests: 5 passed
- Plugin routing and transcript tests: 12 passed
- Agent-server RPC integration test: passed
- Focused TypeScript builds: passed


ts/packages/copilot-macros/src/redaction.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+const secretKey =
+    /^(?:authorization|cookie|password|passwd|secret|token|api[_-]?key|access[_-]?key)$/i;
```
GeorgeNgMsft, line 5: We may want to make this more robust with further normalization (lowercasing and removing non-alphanumeric separators). There are popular credentials that could be added such as : 
`accessToken, refreshToken, clientSecret, privateKey, connectionString, credentials, *Token, *Secret, *Key`. Though you may need to be careful with the *filters since some acceptable fields like **tokenCount** might get blocked.

It would probably be better to work with an allowlist rather than explicit try to redact on a blocklist, but I'm not sure how realistic that is.

ts/packages/copilot-plugin/src/hooks/macro-transcript.ts
```diff
@@ -0,0 +1,232 @@
...
+function matchesPrompt(
+    event: TranscriptEvent,
+    expectedPromptHash?: string,
+): boolean {
+    if (event.type !== "user.message") return false;
+    const content = getString(event.data ?? {}, "content");
+    if (!content) return false;
+    return (
+        !expectedPromptHash ||
+        createHash("sha256").update(content).digest("hex") ===
```
GeorgeNgMsft: This hash is created without redaction, but the prompt hook creates expectedPromptHash with a redacted value `redactTraceValue(input.prompt)` in hook-router.ts

```
        const token = await connection.claimMacroRecording({
            sessionId: input.sessionId,
            cwd: input.cwd,
            promptHash: createHash("sha256")
                .update(redactTraceValue(input.prompt) as string)
                .digest("hex"),
        });
```

But the hash created here is made from the unredacted raw event, and so in a scenario with secrets in the prompt, this match will fail. You should redact here too and maybe consider creating a shared hashing function so all call sites can avoid drift there.

ts/packages/copilot-plugin/src/hooks/hook-history.ts
```diff
@@ -42,106 +46,178 @@ interface TurnSummary {
...
         }
+        if (current === previous) return current;
+        previous = current;
+        await new Promise((resolve) => setTimeout(resolve, intervalMs));
     }
+    return undefined;
+}
+
+async function captureMacroRecording(input: AgentStopInput): Promise<void> {
+    const connection = await connectToAgentServer();
```
GeorgeNgMsft: Do we intentionally want to throw if agentServer connection fails? Otherwise, maybe we should move it into a try/catch with a warning.

GeorgeNgMsft (approved):
Overall looks good to me! Left a couple comments on things to consider in a follow up or here. I think the prompt match behavior should be fixed though.

ts/packages/copilot-plugin/src/hooks/macro-transcript.ts
hillary-mutisya: Updated hash to take normalized content

ts/packages/copilot-plugin/src/hooks/hook-history.ts
hillary-mutisya: Fixed

PR #2891, opened 2026-08-17
Fix Android sample agent schema parse and re-registration failures

jebrans:
### Why

The `mobile-2` sample registers itself as a server-side agent named `androidDevice` and ships its action schema inline. Two separate bugs meant that agent never worked against a live server.

The schema declared `days?: ( | "monday" | ... )[]`. The server's TypeScript schema parser cannot handle a parenthesized type used as an array element:

```
Error parsing schema 'androidDevice': Error parsing alias type SetAlarmAction:
Unhandled type node ParenthesizedType
```

Separately, the server keys client agents by **conversation** rather than by connection, and does not reliably drop the registration when a socket dies abruptly. A phone going to sleep produces a half-open socket (`Software caused connection abort`); the reconnect resumes the same conversation and is rejected:

```
Agent registration failed: App agent 'androidDevice' already exists
```

`handleAndroidDeviceInvoke` refuses every `executeAction` while `isClientAgentRegistered` is false, so a single unclean disconnect left device actions permanently unavailable — each reconnect collided again, and the only recovery was waiting for the server to reap the orphan.

### What this changes

- Extracts a named `AlarmRepeatDay` alias and uses `days?: AlarmRepeatDay[]`. The schema is semantically identical and parses.
- `registerClientAgent` recognises the collision and accepts the existing registration rather than treating it as a failure, so actions keep working across an unclean disconnect.
- The collision is **not** retried. It is an orphaned registration that the server only releases when its own keepalive gives up on the dead connection, which no delay short enough to sit inside a connect will outlast — a retry only postpones the same fallback.
- The fallback gets its own status text (`Connected - reusing existing Android registration`). The guarantee is weaker than a clean registration: until the server drops the old connection it may still route invokes there, and that must not be presented as a normal success.
- The matcher requires the whole `App agent '<name>' already exists` phrase. Only the quoting is loose, since it carries no meaning and is the likeliest part to change; the phrase stays anchored so an agent whose name merely *starts* with this one cannot match, and a genuine failure — a schema parse error, say — still surfaces.
- `webSocket` becomes `@Volatile`. It is assigned on the caller's thread in `connect()` and read from the OkHttp reader thread with no barrier between them.
- `connect()` claims the new connection generation before clearing pending invokes or cancelling the socket, closing a window in which a superseded connection's callback still believed it was current and could write its state over the connection replacing it. `disconnect()` retires the generation for the same reason, so an in-flight registration failing with `"App closed"` no longer reports a registration failure over the final state.

### Tests

14 new cases, all in existing unit-test style — no new tooling.

`AgentAlreadyRegisteredErrorTest` (10) pins the matcher: the collision is recognised through whitespace, casing, surrounding context and any quoting style; a collision reported for a *different* agent, or for one whose name starts with `androidDevice`, is not treated as ours; genuine failures (schema parse error, `Disconnected`, `WebSocket is not connected.`, empty, null) still surface; and the reused status stays distinct from the clean one.

`AndroidDeviceSchemaAssetTest` (4) reads the shipped `.ts` asset off disk and guards the failure modes that are silent at runtime: no parenthesized-union array; every `actionName` literal is handled by `AndroidDeviceAgent.parseExecuteAction`; the `AndroidDeviceAction` union lists every action type exactly once; the seven `AlarmRepeatDay` values are all accepted by `AlarmActionParser`. It asserts its own regexes matched something, so a reformat of the schema cannot turn the comparison into an empty-to-empty one that passes while checking nothing.

### Verification

`:app:testDebugUnitTest` — 162 tests, 0 failures (148 before). `:app:assembleDebug` — BUILD SUCCESSFUL.

Both schema guards were mutation-tested: reintroducing `days?: ( | "monday" | "tuesday" )[]` and renaming `searchNearby` each failed exactly one test, and the asset was restored byte-identically afterwards.

On a Pixel 10 (Android 16, API 36) against a live server: the agent registers and `@config schema androidDevice` reports `on`; `setTimer`, `setAlarm` including repeat `days` (the field whose syntax caused the parse failure), `dialPhoneNumber` and `searchNearby` all execute; the app survives force-stop and an airplane-mode drop.

The collision branch is covered by unit tests only — the server reaps promptly on a clean close, so the orphaned registration could not be reproduced on demand on a live server.

### DIFF
--- a/android/samples/mobile-2/app/build.gradle.kts
+++ b/android/samples/mobile-2/app/build.gradle.kts
@@ -50,6 +50,13 @@ android {
         compose = true
         buildConfig = true
     }
+    testOptions {
+        unitTests {
+            // WebSocketManager logs through android.util.Log, whose unit-test
+            // stub throws unless the stubs return defaults.
+            isReturnDefaultValues = true
+        }
+    }
 }
 
 dependencies {
--- a/android/samples/mobile-2/app/src/main/assets/typeagent/androidDeviceSchema.ts
+++ b/android/samples/mobile-2/app/src/main/assets/typeagent/androidDeviceSchema.ts
@@ -13,6 +13,16 @@ export type AndroidDeviceAction =
     | WebSearchAction
     | OpenWebPageAction;
 
+// A day of the week an alarm can repeat on.
+export type AlarmRepeatDay =
+    | "monday"
+    | "tuesday"
+    | "wednesday"
+    | "thursday"
+    | "friday"
+    | "saturday"
+    | "sunday";
+
 // Sets an alarm on the Android device.
 // Use when the user asks to create or schedule an alarm for a time of day.
 export type SetAlarmAction = {
@@ -24,15 +34,7 @@ export type SetAlarmAction = {
         time: string;
         // Days the alarm repeats on. Omit for a one-off alarm that rings at the
         // next occurrence of the given time.
-        days?: (
-            | "monday"
-            | "tuesday"
-            | "wednesday"
-            | "thursday"
-            | "friday"
-            | "saturday"
-            | "sunday"
-        )[];
+        days?: AlarmRepeatDay[];
     };
 };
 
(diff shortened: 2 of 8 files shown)


GeorgeNgMsft (approved):
lgtm!

PR #2893, opened 2026-08-17
Enforce empty-gold negative fairness at validation

datduyng:
Empty-gold negatives must describe a hard abstention. Apply the deterministic utterance shape gate when a reviewer marks a negative as fair, require the assessment path to match the generated case exactly, and reject mechanical path aliases. Add a nested Git LFS rule for JSONL files in this local benchmark run.

Tested:
- pnpm --dir ts/packages/benchmarks build
- pnpm --dir ts/packages/benchmarks jest-esm --testPathPattern="(negativeFairness|emptyGoldUtterance)\\.spec\\.js" --runInBand


ts/packages/benchmarks/src/translationBench/synthesizer/negativeFairness.ts
```diff
@@ -147,74 +141,12 @@ export function decomposedFlagsAreFair(
...
-): boolean {
-    return (
-        assessment.fairEmptyGold &&
-        FAIR_KINDS.has(assessment.kind) &&
-        decomposedFlagsAreFair(assessment)
-    );
+    ].join(", ");
 }
 
 export function checkTranslationBenchNegativeFairnessAssessment(
```
GeorgeNgMsft, line 152: It seems like a lot of the validation you described in the PR summary is implemented here, but this function doesn't seem to be wired anywhere yet (except the tests). Is that expected or are you missing some commits here?

ts/packages/benchmarks/src/translationBench/synthesizer/negativeFairness.ts
```diff
@@ -224,23 +156,24 @@ export function checkTranslationBenchNegativeFairnessAssessment(
...
             path: assessment.path,
             utterance,
-            reason,
+            reason:
+                assessment.fairEmptyGold && FAIR_KINDS.has(assessment.kind)
+                    ? `inconsistent fair empty-gold flags: ${inconsistentFairFlagsMessage(assessment)}`
+                    : assessment.reason,
         };
     }
+    const shape = assessEmptyGoldUtterance(utterance);
```
GeorgeNgMsft, line 170: **assessEmptyGoldUtterance()** is being used here as a deterministic gate and there are some potential (but probably intentional) gaps there. That gate seems to only recognize splitting on semi colons, dashes, and hyphens. 

https://github.com/microsoft/TypeAgent/blob/e1e14c1a57a3e2d4f9fb8a3f5474498383c3130e/ts/packages/benchmarks/src/translationBench/synthesizer/emptyGoldUtterance.ts#L4

I'm not sure what is in the gold tests right now (so not certain if this will affect your benchmarking) but in a realistic scenario, we will have inputs that will involve periods and commas that merge two clauses (but are treated as one incorrectly currently).

For example : 
```
"Don't open example.com. Close this tab."
"Don't open example.com, but close this tab."
```
These might be accepted as refusals when they aren't. 

There are good reasons why we don't want to naively accept periods and commas, but there are also common rules for things like `, but` and `, then` which are regularly used conjucations that would be worth filtering on. 

That being said, worth noting this is a pretty English-centric cleanup, but I think that is fine as it seems to be the scope at the moment.

ts/packages/benchmarks/src/translationBench/synthesizer/negativeFairness.ts
```diff
@@ -81,58 +80,53 @@ export interface TranslationBenchNegativeFairnessResult {
...
     }
     return byPath;
 }
 
-/** True when the four atomic flags describe a hard abstain with no toolable follow-on. */
+export function translationBenchNegativeAssessmentsJsonSchema(): Record<
+    string,
+    unknown
+> {
+    const schema = z.toJSONSchema(assessmentsSchema) as Record<string, unknown>;
```
GeorgeNgMsft, line 109: This seems to have replaced a manual schema construction that had structure like : 
```
    const pathSchema =
        requiredPaths !== undefined && requiredPaths.length > 0
            ? { type: "string", enum: [...requiredPaths] }
            : { type: "string", minLength: 1 };

and 

      ...(requiredPaths !== undefined
            ? {
                  minItems: requiredPaths.length,
                  maxItems: requiredPaths.length,
              }
            : {}),
```
Do you have any concerns with losing those constraints?

GeorgeNgMsft (approved):
Left a few questions and suggestions. This is fine as an incremental PR but unsure if the new validator changes are left intentionally let unwired.

ts/packages/benchmarks/src/translationBench/synthesizer/negativeFairness.ts
datduyng: We aren't trying to cover 100%. we are trying to generate the 1000 rows that catch 90% of issues. 

Note: this dataset is meant for single utterances. `, but and , then` would get into multi-utterance.

datduyng: closing as this is not applicable to single action dataset. I will follow up with George offline as well

ts/packages/benchmarks/src/translationBench/synthesizer/negativeFairness.ts
datduyng: this is fine for one off data cleaning code

ts/packages/benchmarks/src/translationBench/synthesizer/negativeFairness.ts
datduyng: yes, will be on the next PR. since original pr is large, breaking it up into smaller ones

PR #2897, opened 2026-08-18
Add translation benchmark action eligibility policy loader

datduyng:
## Summary

- add a validated, versioned action eligibility policy
- load the packaged policy with a stable content hash
- expand explicit and onboarding exclusions plus field verification overrides

## Testing

- pnpm --filter @typeagent/benchmarks build
- focused policy loader test
- repository policy check


ts/packages/benchmarks/src/translationBench/policy/action-eligibility.json
```diff
@@ -0,0 +1,445 @@
...
+      "reasons": ["not_user_disambiguable"]
+    },
+    {
+      "type": "action",
+      "id": "utility.claudeTask",
+      "reasons": ["internal_utility"]
+    },
+    {
+      "type": "action",
+      "id": "dispatcher.unknown",
```
GeorgeNgMsft, line 81: This does not seem to exist in the generated catalog https://github.com/microsoft/TypeAgent/blob/main/ts/packages/benchmarks/src/translationBench/catalog.generated.json 

This is being flagged as potentially failing validation of the whole policy because assertRemovedActionsMatchCatalog() may throw as soon as it encounters this.

GeorgeNgMsft (approved):
Please double check my comment but otherwise looks fine

ts/packages/benchmarks/src/translationBench/policy/action-eligibility.json
datduyng: image


It's a bad design that we have dispatcher.unknown as a catch all tbh

PR #2898, opened 2026-08-18
Add translation benchmark ambiguity analysis

datduyng:
- classify agreement across independent translation probes
- enumerate the seed and positive generated cases that require probing
- preserve model order and report missing observations as probe errors

### DIFF
--- a/ts/packages/benchmarks/src/translationBench/synthesizer/ambiguityProbe.ts
+++ b/ts/packages/benchmarks/src/translationBench/synthesizer/ambiguityProbe.ts
@@ -1,0 +1,227 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import type { TranslationBenchBenchmarkAction } from "./benchmark.js";
+import type {
+    TranslationBenchGeneratedCandidate,
+    TranslationBenchReviewIssue,
+} from "./generationCandidate.js";
+
+export interface TranslationBenchAmbiguityProbeAction {
+    schemaName: string;
+    actionName: string;
+    parameters?: Record<string, unknown>;
+}
+
+export interface TranslationBenchAmbiguityProbeObservation {
+    model: string;
+    actions: TranslationBenchAmbiguityProbeAction[];
+    error?: string;
+}
+
+export interface TranslationBenchAmbiguityProbeRequest {
+    model: string;
+    utterance: string;
+    history?: unknown;
+    activeSchemas: readonly string[];
+}
+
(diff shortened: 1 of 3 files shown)


ts/packages/benchmarks/src/translationBench/synthesizer/ambiguityProbe.ts
```diff
@@ -0,0 +1,227 @@
...
+        );
+    }
+    const targets = listTranslationBenchAmbiguityProbeTargets(
+        options.candidate,
+    );
+    const cases: TranslationBenchAmbiguityProbeCaseResult[] = [];
+    for (const target of targets) {
+        const observations = await Promise.all(
+            models.map((model) =>
+                options.translator.translate({
```
GeorgeNgMsft: You store errors in each observation so that you can skip observations with an error right? but the use of Promise.all() here need the translator to handle each exception individually and resolving them. Otherwise, you might reject on the entire operation.

ts/packages/benchmarks/src/translationBench/synthesizer/ambiguityProbe.ts
```diff
@@ -0,0 +1,227 @@
...
+    observations: readonly TranslationBenchAmbiguityProbeObservation[],
+): {
+    agreement: TranslationBenchAmbiguityAgreement;
+    routes: string[];
+} {
+    const gold = goldRouteKey(expected);
+    const okRoutes: string[] = [];
+    for (const obs of observations) {
+        if (obs.error !== undefined && obs.error.trim().length > 0) {
+            continue;
```
GeorgeNgMsft, line 109: Is the skip here intentional? It seems like it can result in gold + error mixes resulting as **unanimous_gold**, maybe consider introducing one more state to better represent the mixed success.

Image

GeorgeNgMsft (approved):
Looks good to me, a few comments

ts/packages/benchmarks/src/translationBench/synthesizer/ambiguityProbe.ts
datduyng: will discuss offlne with George. for now close

ts/packages/benchmarks/src/translationBench/synthesizer/ambiguityProbe.ts
datduyng: Good catch. I added per-model error handling so one rejected translation is recorded as an error observation instead of rejecting the entire probe. I also added a regression test.

PR #2902, opened 2026-08-19
Remove legacy Android WebView sample

jebrans:
### Why

The original `mobile` sample did not provide its own Android chat experience. It loaded the TypeAgent shell renderer inside an Android `WebView` and injected native Android APIs into that page through a JavaScript bridge.

That architecture required Android-specific behavior in the otherwise desktop-oriented shell renderer:

- `main.ts` called `Bridge.interfaces.Android.domReady()` after shell initialization so the Android host could deliver a prompt supplied through an Android intent.
- `speech.ts` detected the injected Android API and delegated one-shot speech recognition to `Bridge.interfaces.Android.recognize()`.
- `lib.android.d.ts` declared the global `Android` and `Bridge` objects injected by the WebView host.

The replacement `mobile-2` sample uses a different architecture. It provides a native Jetpack Compose UI, handles Android functionality in the app, and communicates directly with the TypeAgent agent server over WebSocket. It does not embed the shell renderer or inject JavaScript bridge objects.

Keeping the Android branches in `shell/src/renderer` would therefore preserve support for a host that no longer exists. It would also make the Electron shell appear to depend on Android even though the dependency ran in the opposite direction: the retired Android app depended on the reusable shell renderer.

### What this changes

- Deletes the legacy `android/samples/mobile` WebView-based application.
- Removes the `Bridge.interfaces.Android.domReady()` callback from the shell renderer.
- Removes `getAndroidAPI()` and the renderer's dependency on the injected global `Android` object.
- Removes the Android-specific one-shot and continuous speech-recognition branches from `speech.ts`.
- Simplifies `needSpeechToken()` so only local Whisper bypasses the normal speech-token requirement.
- Deletes `lib.android.d.ts`, which described globals supplied only by the retired WebView host.
- Updates the Android sample index to point to the native `mobile-2` implementation.
- Removes the `.gitignore` entry that applied only to the deleted sample.

The normal Electron and browser-compatible shell paths are unchanged. Speech recognition continues to use local Whisper when enabled and the Microsoft Cognitive Services Speech SDK otherwise.

### Verification

The shell renderer TypeScript check completed successfully:

`tsc -p tsconfig.web.json --noEmit` — 0 errors.

Repository checks confirmed:

- No tracked source references remain for `getAndroidAPI`, `Bridge.interfaces.Android`, `globalThis.Android`, or `lib.android.d.ts`.
- No tracked references remain for the deleted `android/samples/mobile` project.
- `android/samples/mobile-2` and `android/samples/wearos` remain intact.
- `git diff --check` passes.


GeorgeNgMsft (approved):
Lgtm!

PR #2904, opened 2026-08-19
Add automatic Copilot coding workflows to TypeAgent

hillary-mutisya:
Integrates the Copilot SDK coding agent into TypeAgent’s fallback routing. Coding requests are detected automatically, run against authorized local working directories, retain context across follow-ups, and report validation outcomes.
Screenshot 2026-08-18 205833

Includes agent-server path policy, session lifecycle management, permission enforcement, completion tracking, SDK 1.0.9 upgrade, and a deterministic coding smoke-test project. Removes the need for an explicit coding mode command.

Screenshot 2026-08-18 210034

### DIFF
--- a/ts/packages/agentServer/server/src/workingDirectoryPolicy.ts
+++ b/ts/packages/agentServer/server/src/workingDirectoryPolicy.ts
@@ -1,0 +1,83 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import fs from "node:fs";
+import path from "node:path";
+
+export type WorkingDirectoryPolicy = {
+    allowedRoots: string[];
+    defaultRoot?: string;
+};
+
+export type WorkingDirectoryResolution = {
+    workingDirectory?: string;
+    source?: "requested" | "default";
+    rejectedRequested?: boolean;
+};
+
+function canonicalDirectory(value: string): string | undefined {
+    try {
+        const canonical = fs.realpathSync(path.resolve(value));
+        return fs.statSync(canonical).isDirectory() ? canonical : undefined;
+    } catch {
+        return undefined;
+    }
+}
+
+function isWithinRoot(candidate: string, root: string): boolean {
+    const relative = path.relative(root, candidate);
+    return (
+        relative === "" ||
+        (!relative.startsWith("..") && !path.isAbsolute(relative))
+    );
+}
+
+export function loadWorkingDirectoryPolicy(
+    env: NodeJS.ProcessEnv = process.env,
+): WorkingDirectoryPolicy {
+    const allowedRoots = (env.TYPEAGENT_CODE_ALLOWED_ROOTS ?? "")
+        .split(path.delimiter)
+        .map((entry) => entry.trim())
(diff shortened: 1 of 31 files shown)


jebrans (approved):

ts/packages/agentServer/server/src/workingDirectoryPolicy.ts
```diff
@@ -0,0 +1,130 @@
...
+export function resolveWorkingDirectory(
+    requested: string | undefined,
+    policy: WorkingDirectoryPolicy,
+): WorkingDirectoryResolution {
+    const allowedRoots = policy.allowedRoots
+        .map(canonicalDirectory)
+        .filter((root): root is string => root !== undefined);
+    const allowed = (candidate: string): boolean =>
+        policy.allowedRoots.length === 0 ||
+        allowedRoots.some((root) => isWithinRoot(candidate, root));
```
GeorgeNgMsft, line 106: If every path is invalid, it will get filtered into an empty list and treated the same as an open policy here. However, a list of invalid paths is not the same as an intentional empty list, as it would indicate that a security policy was intentionally placed here. We should probably check if the configured paths is non-zero along with the allowedRoots.length === 0.

ts/packages/dispatcher/dispatcher/src/reasoning/codingCompletion.ts
```diff
@@ -0,0 +1,134 @@
...
+    let stopBlockIssued = false;
+
+    return {
+        onToolStart(toolName: string, args: unknown): void {
+            if (isValidationToolUse(toolName, args)) {
+                validationAttempted = true;
+            }
+        },
+        onToolSuccess(toolName: string, args: unknown): void {
+            if (isWriteToolUse(toolName, args)) {
```
GeorgeNgMsft, line 80: A turn can have multiple edits, so we need to reset the validation state after each write() call.

ts/packages/dispatcher/dispatcher/src/reasoning/codingRouting.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+export function isCodeAgentRequest(requestAction: RequestAction): boolean {
+    return requestAction.actions.some(({ action }) =>
+        action.schemaName.startsWith("code"),
+    );
+}
+
+const ACTION_PATTERN =
+    /\b(add|build|compile|create|debug|delete|edit|fix|format|implement|lint|migrate|modify|refactor|remove|rename|run|test|typecheck|update|write)\b/i;
```
GeorgeNgMsft, line 47: ACTION_PATTERN is a superset of MUTATION_REQUEST_PATTERN in codingCompletion, so to avoid drift, maybe we should break this up into NON_MUTATION_REQUEST and MUTATION_REQUEST_PATTERN and use a combination of these two patterns to represent ACTION_PATTERN. It trades off a bit of complexity to avoid drift possibility. We should probably add "Change" to the list too.

Small concern, but worth considering

GeorgeNgMsft:
This are smaller issues, I can address them in a follow up pr

PR #2907, opened 2026-08-19
Parallelize translation benchmark generation

datduyng:
Parallelize generation while preserving unique utterances and serialized checkpoints.


ts/packages/benchmarks/src/translationBench/synthesizer/datasetGenerator.ts
```diff
@@ -245,8 +252,41 @@ export function createTranslationBenchGenerationSchedule(
...
+            ids.push(id);
+            idsByActionName.set(actionName, ids);
+        }
+        for (const ids of idsByActionName.values()) {
+            if (ids.length > 1) {
+                for (const id of ids) ambiguousActionIds.add(id);
+            }
+        }
+    }
+    const excludedActionIds =
```
GeorgeNgMsft: It seems that the scheduler here excludes duplicate-named actions but the validator does not make these same exclusions in the benchmark. The validator would then expect more actions than the scheduler adds, so that will likely cause an exception to be thrown.

Both the scheduler and validator should unify their exclusion logic (though you should decide on the exact policy for that). 

---

Scheduler pays attention to ambiguousActionIds but Validator only seems to care about packaged excluded actions.

https://github.com/microsoft/TypeAgent/blob/56f9e04f82c48ef57d34828be538c31647b5b24f/ts/packages/benchmarks/src/translationBench/synthesizer/benchmark.ts#L2203

ts/packages/benchmarks/src/translationBench/synthesizer/datasetGenerator.ts
```diff
@@ -245,8 +252,41 @@ export function createTranslationBenchGenerationSchedule(
...
+    {
+        const idsByActionName = new Map<string, string[]>();
+        for (const key of census.qualifiedActionKeys) {
+            const [schemaName, actionName] = JSON.parse(key) as [
+                string,
+                string,
+            ];
+            const id = `${schemaName}.${actionName}`;
+            if (baseExcludedActionIds.has(id)) continue;
+            const ids = idsByActionName.get(actionName) ?? [];
```
GeorgeNgMsft: I see that you decide on ambiguous actions if they share the same action name, but scenarios like :

**localPlayer.stop** and **visualStudio.stop** are not the same. They share an action name but are not duplicates. Are you sure we want to exclude these examples for now? It seems like a valid scenario that is a ignored at the moment.

GeorgeNgMsft (approved):
The ambiguous actions isn't a blocking issue, but the scheduler / validator exclusion logic mismatch could be considered one. Please evaluate that and address it before check in. Otherwise looks good!

ts/packages/benchmarks/src/translationBench/synthesizer/datasetGenerator.ts
datduyng: Fixed in d8950717a. The scheduler and validator now use the same packaged qualified-action exclusions.

ts/packages/benchmarks/src/translationBench/synthesizer/datasetGenerator.ts
datduyng: Agreed. A shared bare name does not make qualified actions duplicates, so both actions remain eligible.

PR #2908, opened 2026-08-19
Require zero-action negative fairness

datduyng:
Require negative fairness assessments to prove hard abstention and canonical case paths.


jebrans (approved):

ts/packages/benchmarks/src/translationBench/synthesizer/datasetGenerator.ts
```diff
@@ -745,6 +786,18 @@ export function finalizeTranslationBenchGeneratedCaseLineage(
...
             );
+        // Attach deterministic soft-match specs so the runner does not exact-
+        // match free-text params (e.g. originalRequest). Derived from the
+        // packaged grader; excluded from the canonical payload hash above.
+        const specs = parameterScoreSpecsForExpectedActions(
+            grader,
+            probe.expectedActions,
+        );
+        if (hasUsableParameterScoreSpecs(specs)) {
+            probe.parameterScore = specs;
```
GeorgeNgMsft, line 797: It seems like there might be a small potential gap between the authored schema source and generated catalog where drift can occur. If you run pnpm run gen-catalog on a stale .pas.json file, then it'll result in a gap between the scriptActions.mts and the *.pas.json. 

When that gap occurs, unexpected action gets marked as undefined and it will delete the parameterScore resulting in the following flow from this recent change : 

Live action exists
→ packaged grader entry is missing
→ helper returns [undefined]
→ hasUsableParameterScoreSpecs() returns false
→ parameterScore is deleted
→ runner defaults to exact matching

This can mess up the scoring for that action. At the time of this comment, it seems there's a drift of 6 actions in the latest atalog.generated.json (commited 8/17) missing some powershell actions before. The simple solution is probably just folding a full pnpm run build into gen-catalog.

datduyng: Discussed offline. 

```
Live action exists
→ packaged grader entry is missing
→ helper returns [undefined]
→ hasUsableParameterScoreSpecs() returns false
→ parameterScore is deleted
→ runner defaults to exact matching
```

this scenario cannot happen because no expectAction can that doesnt have a byAction entry but will fix the dependency issue on the next PR as it's an additional scope for the PR. Thank you for thoughtful review :)

PR #2909, opened 2026-08-19
Validate negative review assessment paths

datduyng:
Validate negative review coverage and reject missing or mismatched assessment paths.


GeorgeNgMsft (approved):
Same scheduler vs validator action eligibility concern as in https://github.com/microsoft/TypeAgent/pull/2907 but otherwise seems good

datduyng:
Addressed in 6cff12c61. Scheduler and validator now use the same packaged qualified-action exclusions.

PR #2914, opened 2026-08-20
Support multiple android devices connected to Agent server

jebrans:
### Target design

```
                 conversation "Kitchen"
                 +--------------------------+
   dispatcher -> | agent "androidDevice"    |   <- ONE schema, registered once
                 |  (multiplexer)           |
                 +-----------+--------------+
                             | picks one
              +--------------+--------------+
              v              v              v
        Pixel 8         Galaxy Tab      Pixel Fold
      instanceId=a    instanceId=b    instanceId=c
      connection 1    connection 2    connection 4
```

Title: Let several devices register one client-hosted agent

---

### Why

Every Android device registers a client-hosted agent under the same fixed name
`androidDevice` (`AndroidDeviceAgent.NAME`). Agent names are unique per
conversation — so the second device is rejected in `appAgentManager.ts:1054`:
```
Agent registration failed: App agent 'androidDevice' already exists
```

### What this changes

One logical agent, many device instances behind a multiplexer, reference
counted. The schema is registered with the dispatcher exactly once; each device
keeps its own rpc proxy; each call goes to one device. This is the pattern the
browser agent already uses (`agentWebSocketServer.mts`), not a new one.

- New `clientAgentRegistry.ts` owns the groups. Registration is join-or-create:
  the first device creates the group and calls `addDynamicAgent`, later devices
  with the same schema just add an instance.
- **Sharing a name is opt-in per registration, so no existing client changes
  behaviour.** The shell registers a client-hosted agent through this same path
  (`instance.ts:304`) and expects to be the only host of `@shell`. It does not
  pass `multiInstance`, so a second shell is still rejected with the original
  error, exactly as today. The Android app opts in. The browser extension and
  VS Code never reach this code — their `registerClientAgent` is
  `notSupported`. The client that creates the registration decides, so nobody
- Sharing is opt-in per registration, and the group's creator decides. The shell
  registers without opting in and keeps exactly today's behaviour; only the
  Android app asks to share a name. A client that opts in cannot join a group
  whose creator did not, so nobody can widen someone else's agent.
  A reconnecting `instanceId` replaces in place either way, since that path
  fails today and so cannot regress anything.
- Android sample: a `DeviceIdentity` seam backed by a `SharedPreferences` UUID
  (never a hardware identifier, so no new permissions) with `Build.MODEL` as the
  default display name. It is injected rather than built from a `Context`
  because the sample's unit tests are plain JVM JUnit and
  `RegistrationCollisionRecoveryTest` drives a real `WebSocketManager`. The
  parameter is required, not defaulted, so the production path cannot silently
  fall back to a throwaway identity and make every reconnect look like a new
  device.
- `clientType` was `"extension"`, copied from the browser extension. Nothing
  reads it today (`getBrowserControlForRequest` is defined but never called), so
  this is a correctness fix rather than a behaviour change. It ships with the
  protocol change, since `"android"` is rejected until the union is widened.

### Tests

29 new cases in the agent-server package.

`clientAgentRegistry.spec.ts` (27) covers the behaviour against the real
registry: a second device joins and the dispatcher still holds one agent; a
repeated `instanceId` replaces rather than duplicates; a different schema is
rejected with the version message while the same schema with reordered manifest
keys is accepted (the JSON key-order trap); device A disconnecting leaves B
working; `removeDynamicAgent` fires exactly once and a double removal is inert;
a name-only unregister removes nothing when the caller owns nothing; routing
picks the requester, falls back to the only device, and otherwise throws while
running nothing anywhere; asking from one device does **not** make it the target
for a later request from elsewhere (the guard against silent stickiness); two
devices sharing a display name are listed distinctly in the error; a read-only
call still runs when the target is ambiguous; overlapping requests from two
connections land on their own devices; two registrations issued before either
resolves produce one group and one `addDynamicAgent`, plus the mirror add/remove
race; a client that does **not** opt in staying the only host of its agent,
including a client that does opt in being unable to join its group — the shell
case, which also pins the original `already exists` text and that a known
`instanceId` still reconnects; and the sleeping-phone case, where the late
disconnect of a stale socket leaves the reconnected device registered.

`clientAgentIntegration.spec.ts` (2) drives two real client connections through
the real connection handler and agent-rpc transport over an in-memory channel
pair: both register `androidDevice`, each executes its own actions, A's
disconnect leaves B working, and the last one out tears the dynamic agent down
once. The second case pins backward compatibility — a client that sends no
identity registers as a single instance keyed by its connection.

The routing tests use a `SessionContext` whose `popupQuestion` throws, so a
future change that reintroduces a prompt fails the suite rather than quietly
interrupting the user.

`sessionContext.spec.ts` gains one case: `currentConnectionId` is the requesting
connection for a client request and `undefined` for a server-initiated one.

Android: 3 cases pinning that registration carries the device identity, that it
opts in with `multiInstance`, and that the eviction call names no instance,
since giving it one would reintroduce a way to drop a live registration.

### Verification

`pnpm run clean && pnpm run build` — BUILD SUCCESSFUL.
`pnpm --filter agent-server test` — 66 passed (48 before).
`pnpm --filter agent-dispatcher test:local` — `sessionContext.spec` 16 passed.
`pnpm run prettier:changed:fix` — clean.
`:app:testDebugUnitTest` — 171 tests, 0 failures (169 before).
`:app:assembleDebug` — BUILD SUCCESSFUL.


GeorgeNgMsft (approved):
PR looks good, I think the architecture makes sense given how different the context of working on android is (agents living on the device separately from where agent-server is hosted). 

A challenge I think might be worth keeping in mind is that the current multi-instance agents design works well as stateless agents. If we do eventually find scenarios where we want stateful agents that live on android, then there are some lifecycle gaps that should be addressed here when compared to the existing AppAgent which does support stateful-ness. 

A scan of the differences currently highlights : 
- **Reconnected devices retain invalid RPC context IDs**

`ts/packages/agentServer/server/src/clientAgentRegistry.ts:564-573` replaces the proxy but preserves contexts owned by the previous RPC server. After reconnect, calls can fail with `Invalid agent context ID`. Reinitialize context state before routing to the replacement proxy.

- **Failed initialization still registers a broken device**

`clientAgentRegistry.ts:488-514,612-617` catches initialization failures and returns registration success. The unusable device remains selectable for requests. Initialization should fail registration and remove or quarantine the instance, with partial lifecycle work cleaned up.

- **Instance removal bypasses lifecycle teardown**

`clientAgentRegistry.ts:659-660,736-744` deletes the proxy/context before teardown. With other devices remaining, teardown is never invoked; for the final device, `removeDynamicAgent` sees an empty mux. Explicit unregister can therefore leave background tasks, ports, or context resources active. Teardown must run while the instance remains addressable.

These issues would become observable in a lifecycle-aware agent in these scenarios 
```
initialize → context allocated remotely
reconnect → cached context ID becomes invalid

late join → initialization fails
          → device nevertheless remains routable

unregister → background work and context are never closed
```

I would recommend either : 
- Fix them now because `multiInstance` is exposed as a generic `AppAgent` capability.
- Explicitly constrain `multiInstance` to context-free agents in this PR, then add full lifecycle support separately.

GeorgeNgMsft:
After this PR, I would consider evaluating Agent-Server's architecture for multi-client support and for elements to adopt/or change to refine our standard for similar scenarios moving forward. Some things worth considering : 

- **Conversation-namespaced agent RPC channels:** Prevent the same agent name in multiple conversations on one connection from colliding or disrupting another conversation.

- **Consolidated connection/device records and explicit readiness states:** Keep identity, connection ownership, proxy, context, and health synchronized; route only to fully initialized devices.

- **Action-level targeting by stable device ID:** Let shell or cross-device requests explicitly select a device instead of failing whenever multiple devices are connected.

- **Explicit routing metadata instead of method-name classification:** Avoid silently misrouting new or unusual `AppAgent` operations because a hard-coded list incorrectly assumes whether they have side effects.

- **Better health diagnostics and device enumeration:** Let users and operators see which devices are connected, ready, reconnecting, or failed and understand why routing is unavailable.

- **Centralized, idempotent cleanup orchestration:** Ensure unregister, disconnect, reconnect, conversation leave, and shutdown all release resources consistently without double-cleanup races or leaks.

PR #2925, opened 2026-08-21
Harden automatic coding workflow validation

GeorgeNgMsft:
Follow-up to #2904 addressing review findings in the automatic coding workflow.

## Summary
- Fail agent-server startup when `TYPEAGENT_CODE_ALLOWED_ROOTS` contains nonempty entries but none resolve to valid directories, instead of silently falling back to unrestricted access. Empty, whitespace-only, or delimiter-only configuration remains intentionally unrestricted.
- Invalidate successful validation evidence after every subsequent successful or failed write-capable tool so completion reflects the final file state, including possible partial writes.
- Share mutation intent classification between coding routing and completion, including common `change` requests.
- Add focused regression coverage for each behavior.

## Validation
- Built `agent-dispatcher` and `agent-server`.
- Ran focused coding completion, coding routing, and working-directory policy tests.
- Passed complexity and lint ratchets against `origin/main`.

### DIFF
--- a/ts/packages/agentServer/server/src/workingDirectoryPolicy.ts
+++ b/ts/packages/agentServer/server/src/workingDirectoryPolicy.ts
@@ -79,18 +79,14 @@ function isWithinRoot(candidate: string, root: string): boolean {
 export function loadWorkingDirectoryPolicy(
     env: NodeJS.ProcessEnv = process.env,
 ): WorkingDirectoryPolicy {
-    const configuredAllowedRoots = env.TYPEAGENT_CODE_ALLOWED_ROOTS;
-    const configuredRoots = (configuredAllowedRoots ?? "")
+    const configuredRoots = (env.TYPEAGENT_CODE_ALLOWED_ROOTS ?? "")
         .split(path.delimiter)
         .map((entry) => entry.trim())
         .filter(Boolean);
     const allowedRoots = configuredRoots
         .map(canonicalDirectory)
         .filter((entry): entry is string => entry !== undefined);
-    if (
-        configuredAllowedRoots !== undefined &&
-        (configuredRoots.length === 0 || allowedRoots.length === 0)
-    ) {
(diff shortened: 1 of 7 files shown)


ts/packages/agentServer/server/src/workingDirectoryPolicy.ts
```diff
@@ -79,12 +79,18 @@ function isWithinRoot(candidate: string, root: string): boolean {
...
-    const allowedRoots = (env.TYPEAGENT_CODE_ALLOWED_ROOTS ?? "")
+    const configuredRoots = (env.TYPEAGENT_CODE_ALLOWED_ROOTS ?? "")
         .split(path.delimiter)
         .map((entry) => entry.trim())
-        .filter(Boolean)
+        .filter(Boolean);
+    const allowedRoots = configuredRoots
         .map(canonicalDirectory)
         .filter((entry): entry is string => entry !== undefined);
+    if (configuredRoots.length > 0 && allowedRoots.length === 0) {
```
jebrans, line 89: This will fire if all the directories are bad, so 1/3 will pass. Is that correct?

jebrans (approved):

ts/packages/agentServer/server/src/workingDirectoryPolicy.ts
GeorgeNgMsft: Yes, this intentional. Previously, it would allow on any paths being configured (regardless of if they were valid or invalid). However, this would leave a weird scenario where a configuration with all invalid paths would be treated as allowing any paths (like a wildcard) after the bad paths get discarded. But we know the user intent is to lock it down, but they may have made mistakes and provided invalid paths. 

This change helps differentiate between intentionally open path list [] versus a path list that becomes intentionally empty (after filtering) like ["test", "test"]. A scenario with 1/3 valid paths should still be valid here as we still have paths in the allowlist.

PR #2929, opened 2026-08-25
Add isolated local MSI builds and payload cleanup diagnostics

curtisman:
## Summary
- add a `build:msi:local` workflow that creates an isolated staging workspace before producing the Windows installer
- make the agent-server deploy and bundle scripts accept explicit pnpm/state locations so local packaging does not depend on the source checkout's linked `node_modules`
- improve payload cleanup failures by logging the underlying error, identifying processes that hold payload files open, and providing recovery guidance
- update the WiX payload extraction helper and documentation for the local build flow

## Details
`build-msi-local.mjs` stages the required workspace files, performs an isolated pnpm install/deploy, builds the agent server and MSI, and verifies the resulting package. Shared scripts gained optional configuration while preserving their existing defaults.

During an MSI upgrade, payload cleanup failures now produce actionable diagnostics in the installer log. The extraction helper records the exception, scans process modules for files loaded from the payload directory, reports matching process names and IDs, and advises the user to close TypeAgent or restart Windows before retrying.

### DIFF
--- a/ts/tools/installers/wix/README.md
+++ b/ts/tools/installers/wix/README.md
@@ -107,14 +107,35 @@ pnpm run build:msi:local -- --skip-build --version 0.0.1-local
 
 The equivalent individual steps are below for troubleshooting.
 
-#### 1. Build the workspace
+#### 1. Verify dependencies and build the workspace
 
 ```powershell
 cd D:\repos\TypeAgent\ts
+pnpm install --prod=false --frozen-lockfile
 pnpm run build
 ```
 
-#### 2. Stage agent-server
+The install command ensures the workspace has development dependencies such as
+`vsce`. It also repairs dependency state left by an interrupted production
+deploy. Omit `pnpm run build` when compiled outputs are already current (the
+equivalent of `--skip-build`).
+
+#### 2. Package VS Code Chat and stage the Copilot plugin
+
+Package tools that require development dependencies before staging the
+agent-server:
+
+```powershell
+pnpm --filter vscode-chat run package
+
+node tools/scripts/stageCopilotPlugin.mjs `
+  --out "$env:TEMP\typeagent-msi-stage\copilot-plugin"
+```
+
+`stageCopilotPlugin.mjs` copies only the runtime files used by the installer and
+writes `bundle-manifest.json`; do not copy the entire plugin `dist` directory.
+
+#### 3. Stage agent-server
 
 The MSI uses `bundleAgentServer.mjs`, matching the published CI artifact. It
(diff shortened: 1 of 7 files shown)


ts/tools/installers/wix/README.md
```diff
@@ -93,9 +116,16 @@ pnpm run build
...
 
+The MSI uses `bundleAgentServer.mjs`, matching the published CI artifact. It
+bundles server entry points and profile agents to minimize files and installed
+size. `deployAgentServer.mjs` is the unbundled variant: it preserves workspace
+package boundaries and a conventional production `node_modules` for debugging
+a repo-less deployment, but produces a substantially larger artifact and is not
+intended for MSI packaging.
+
 ```powershell
 # From D:\repos\TypeAgent\ts
```
GeorgeNgMsft, line 148: There's a slight difference between the manual steps here and what actually runs in the build:msi:local flow. 

Primarily, the protecting the local checkout and development state : 
- Package the VSIX before bundling
- Pass a temporary pnpm state directory
- Restore dependencies afterward
- Pass `--skip-shell-feed-resolution` to `build-msi.mjs` further below so that the flow is fully local (as expected) and avoids checking azure for the shell artifact version

The 1-1 reproduction of the flow also helps avoid any unexpected errors that could arise from any difference in operations

ts/tools/installers/wix/extract-payload.ps1
```diff
@@ -46,6 +46,40 @@ function Write-Log([string]$message) {
     }
 }
 
+function Write-PayloadCleanupDiagnostics([string]$target, [System.Exception]$exception) {
+    Write-Log "ERROR: Unable to clear payload directory '$target'."
+    Write-Log "ERROR: $($exception.GetType().FullName): $($exception.Message)"
+
```
GeorgeNgMsft, line 52: You may want to defensively normalize $target with something like 
`$target = [System.IO.Path]::GetFullPath($target)`

This is because the WiX configuration is passing in a trailing period as part of the root (which is apparently reasonably justifiable). But that trailing period gets expanded into an additional directory `"C:\{directory}\."` eventually gets joined with another directory into `C:\{dir1}\.\{dir2}` before being passed in here.

This will fail the comparison to the $module.Filename below which uses a normalized absolute path from Windows ie `C:\{dir1}\{dir2}`

This is a niche scenario which can only occur during an upgrade via the MSI over a running installation, but the defensive change seems sound.

GeorgeNgMsft (approved):
Minor suggestions, otherwise looks good!

PR #2931, opened 2026-08-26
[OTEL] Fix debug log propagation for app-agents

GeorgeNgMsft:
**Summary**

App-agent command exceptions were caught and displayed inside the execution layer but not propagated to the dispatcher’s canonical failure handler. As a result, failures such as Spotify authentication errors depended on optional  @trace  debug output and were missing structured  command:exception  telemetry. After this fix, app-agent command exceptions are properly propagated.

• Propagate app-agent command exceptions to the dispatcher command boundary.
• Emit structured failure telemetry independently of debug namespace settings.
• Preserve the originating agent when displaying errors.
• Mark execution failures as potentially side-effecting while keeping parse failures safe to retry.
• Add regression coverage for throwing command handlers and correct an existing malformed test fixture.
• Document how to enable agent debug logs in local telemetry.

### DIFF
--- a/ts/packages/dispatcher/dispatcher/src/execute/actionHandlers.ts
+++ b/ts/packages/dispatcher/dispatcher/src/execute/actionHandlers.ts
@@ -1145,12 +1145,6 @@ export async function executeCommand(
                     "AbortError",
                 );
             }
-            const errorDisplay = getErrorDisplayContent(e);
-            if (errorDisplay !== undefined) {
-                actionContext.actionIO.appendDisplay(errorDisplay, "block");
-            } else {
-                displayError(`ERROR: ${e.message}`, actionContext);
-            }
             debugCommandExecError("command execution exception", {
                 requestId: getRequestId(context).requestId,
                 agent: appAgentName,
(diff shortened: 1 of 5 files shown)


ts/packages/dispatcher/dispatcher/src/execute/actionHandlers.ts
```diff
@@ -1145,12 +1145,6 @@ export async function executeCommand(
                     "AbortError",
                 );
             }
-            const errorDisplay = getErrorDisplayContent(e);
-            if (errorDisplay !== undefined) {
-                actionContext.actionIO.appendDisplay(errorDisplay, "block");
```
datduyng, line 1150: Why are we removing this?

datduyng (approved):
Approve because this change has no bebavior impact. Left a question about why we are removing the IO error

ts/packages/dispatcher/dispatcher/src/execute/actionHandlers.ts
GeorgeNgMsft: It’s because the error was being swallowed at this level when it needs to be propagated further to be logged as part of the dispatcher pipeline.

PR #2932, opened 2026-08-26
[VS Code Shell] Update permissions request UX

GeorgeNgMsft:
## Summary

Permission requests now appear in a popup above the chat instead of interrupting the conversation. Users can approve a tool once, for the current request, or for the session. The new `@allow all` and `@allow off` commands, along with natural-language equivalents, control session permissions. A shared reasoning permission policy has also been introduced to unify the behaviors across reasoning providers (like Claude or Copilot). 

- Copilot and Claude reasoning providers use the same host permission policy and popup.
- Multiple requests are queued in one popup and can be viewed with the arrow controls.
- Choosing an option only updates the **Allow** button; nothing runs until the button is clicked.
- `@allow all` allows eligible tools for either reasoning provider in the current TypeAgent session. `@allow off` resets those session permissions.
- Prompts show what the provider wants to run. Safety, sandbox, and policy-controlled actions still require explicit approval.
- Pending prompts recover after reconnects and are cleared when switching conversations.
- The granted permission is shown as a short note on the final response instead of another chat message.

Normal clarification and semantic confirmation questions still appear in the chat and are not bypassed by `@allow all`.

## Screenshots

Permission request popup

Permission scope options

System-authored questions continue to appear in the chat:

Clarification question in chat


datduyng (approved):
Assume George has end to end verified this.

PR #2936, opened 2026-08-26
[VS Code Shell] Add VS Code Shell extension option to MSI

GeorgeNgMsft:
## Summary
Updates the MSI installer to add an option to install the VS Code Shell extension. Also adds ownership keys to correlate with each installation (Chat vs Shell) so we can eventually separate uninstall for both. 

- add an interactive MSI checkbox and `VSCODESHELL` property for installing the TypeAgent VS Code Shell extension
- reuse the VS Code extension installer with separate ownership metadata and no desktop shortcut
- package, publish, and pass the Shell VSIX through local and release MSI builds
- add installer wiring tests to the regular local test command

image

## Validation
- `node --test tools/scripts/test/*.spec.mjs`
- `pnpm run build vscode-shell`
- `npm run package` in `packages/vscode-shell`
- `pnpm run code-complexity:ci`
- `pnpm run code-lint -- --ratchet --base origin/main`
- PowerShell parser validation for `install-vscode-chat.ps1`
- Prettier checks for changed supported files


ts/tools/installers/common/install-vscode-typeagent.ps1
```diff
@@ -3,12 +3,18 @@
 
 <#
 .SYNOPSIS
-  Installs or removes the TypeAgent VS Code Chat extension and desktop shortcut.
+  Installs or removes a TypeAgent VS Code extension (chat or shell) and its
```
jebrans, line 6: Not a bug, but we can rename the file since it is now handling shell too

ts/tools/installers/wix/TypeAgent-AgentServer.wxs
```diff
@@ -618,7 +671,8 @@
       <Custom Action="RegisterCopilotPlugin"    After="MainProgress3">NOT REMOVE~="ALL"</Custom>
       <Custom Action="MainProgress4"            After="RegisterCopilotPlugin">NOT REMOVE~="ALL"</Custom>
       <Custom Action="InstallVsCodeChat"        After="MainProgress4">(NOT REMOVE~="ALL") AND (VSCODECHAT="1")</Custom>
-      <Custom Action="MainProgress5"            After="InstallVsCodeChat">NOT REMOVE~="ALL"</Custom>
+      <Custom Action="InstallVsCodeShell"       After="InstallVsCodeChat">(NOT Installed) AND (VSCODESHELL="1")</Custom>
```
jebrans, line 675: Why not make it same as InstallVsCodeChat 
`(NOT REMOVE~="ALL") `

ts/tools/installers/wix/TypeAgent-AgentServer.wxs
```diff
@@ -371,6 +391,39 @@
...
+                  BinaryKey="WixCA"
+                  DllEntry="WixQuietExec64"
+                  Execute="deferred"
+                  Return="ignore"
+                  Impersonate="yes" />
+
+    <SetProperty Id="UninstallVsCodeShell"
+           Before="UninstallVsCodeShell"
+           Sequence="execute"
+                 Value="&quot;[WindowsFolder]System32\WindowsPowerShell\v1.0\powershell.exe&quot; -ExecutionPolicy Bypass -NoProfile -NonInteractive -File &quot;[TYPEAGENTROOT]install-vscode-chat.ps1&quot; -Action Uninstall -Owner msi -ExtensionId &quot;typeagent.vscode-shell&quot; -OwnershipKey &quot;HKCU:\Software\Microsoft\TypeAgent\VSCodeShell&quot; -NoShortcut -LogPath &quot;[LocalAppDataFolder]TypeAgent\logs\vscode-shell-install.log&quot;" />
```
jebrans: -NoShortcut  is a no-op on the Uninstall part

jebrans (approved):

ts/tools/installers/wix/TypeAgent-AgentServer.wxs
GeorgeNgMsft: This is kind of leftover from the original implementation when Chat was the only experience. NOT REMOVE ~="ALL" lets the repair operation on the MSI installer also act on it. But "NOT Installed" only lets it run the install when it's not installed. 

So the current config should be appropriate as I don't think we want to always run install during repair on Shell (since it's optional and the user selection state does not persist). It's probably better to leave it unhandled on the repair flow.

A follow up flow for managing chat and shell separately maybe needed, but I think eventually we may narrow down to one experience or the other. So, I don't think it's necessary to address right now. For now, we just want the extension easily accessible and installed.

PR #2937, opened 2026-08-26
Add DroidCall dataset analysis command

datduyng:
Adds a TypeScript command that verifies and analyzes the pinned DroidCall dataset from #2831.

```console
$ cd ts && pnpm --filter @typeagent/benchmarks test
Test Suites: 26 passed, 26 total
Tests:       204 passed, 204 total
Result: 204 tests passed, 0 failed
```


ts/packages/benchmarks/src/translationBench/public_datasets/DroidCall/analyze.ts
```diff
@@ -0,0 +1,314 @@
...
+
+const RESULT_REFERENCE = /^#(\d+)$/;
+const DOWNLOAD_OPTION = "--download";
+
+function hasPriorResultReference(
+    value: unknown,
+    priorResultIds: ReadonlySet<number>,
+): boolean {
+    if (typeof value === "string") {
+        const match = RESULT_REFERENCE.exec(value);
```
GeorgeNgMsft: Double check the pattern here as it seems to have a strict comparison to only the value "#N" as a pattern, this can cause the nested training examples to be undercounted.

```
"#0"                              // matches
"directions from downtown to #0" // does not match
"Contact information: #0"        // does not match
```

To fix this, consider searching for every #N reference inside each string and confirm that it matches an earlier call
```
      return [...value.matchAll(RESULT_REFERENCE)].some((match) =>
          priorResultIds.has(Number(match[1])),
      );
```

GeorgeNgMsft (approved):
Minor comment, otherwise looks good!

ts/packages/benchmarks/src/translationBench/public_datasets/DroidCall/analyze.ts
datduyng: Updated the matcher to scan each string for all #N references while still requiring an earlier result ID. The pinned dataset now reports 1,150 nested rows.

PR #2943, opened 2026-08-27
Add DroidCall raw action restoration

datduyng:
Restores provider actions hidden by dispatcher finalization so DroidCall scores the model output instead of `pendingRequestAction` placeholders.

```console
$ # Raw model action: findContact; finalized action: pendingRequestAction
$ # Expected: recover findContact and ignore retries with different parameters
$ pnpm --filter @typeagent/benchmarks run jest-esm --testPathPattern="translationBench.droidCallRawActions.spec.js"
PASS dist/test/translationBench.droidCallRawActions.spec.js
Tests: 1 passed, 0 failed
```

### DIFF
--- a/ts/packages/benchmarks/src/translationBench/public_datasets/DroidCall/eval/droidCallRawActions.ts
+++ b/ts/packages/benchmarks/src/translationBench/public_datasets/DroidCall/eval/droidCallRawActions.ts
@@ -295,8 +295,15 @@ function restoreActions(
         }
         if (
             hasSameActionNames(accepted, raw.finalizedActionNames) &&
-            (translation.rawChosenActions === undefined ||
-                hasSameRawActions(translation.rawChosenActions, raw.actions))
+            hasSameRawActions(
+                translation.rawChosenActions ??
+                    translation.chosenActions.filter(
+                        (action) =>
+                            action.actionName !== PENDING_ACTION_NAME &&
+                            action.parameters !== undefined,
+                    ),
+                raw.actions,
+            )
         ) {
             return raw.actions;
         }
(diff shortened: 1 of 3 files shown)


ts/packages/benchmarks/src/translationBench/public_datasets/DroidCall/eval/droidCallRawActions.ts
```diff
@@ -0,0 +1,312 @@
...
+            raw = parseDroidCallRawResponse(responses[index]!);
+        } catch {
+            continue;
+        }
+        if (
+            hasSameActionNames(accepted, raw.finalizedActionNames) &&
+            (translation.rawChosenActions === undefined ||
+                hasSameRawActions(translation.rawChosenActions, raw.actions))
+        ) {
+            return raw.actions;
```
GeorgeNgMsft, line 308: When several attempts call the same tool with different inputs, the wrong action maybe returned because the responses are matched on name without a parameter comparison. 

Given : 
```
translation = {
    chosenActions: [
        { actionName: "findContact", parameters: { name: "Ada" } },
    ],
    // rawChosenActions is absent
};

responses = [
    '{"actionName":"findContact","parameters":{"name":"Ada"}}',
    '{"actionName":"findContact","parameters":{"name":"Grace"}}',
];
```

The logic searches newest first and reaches "Grace" instead of Ada (from the chosenActions).

GeorgeNgMsft (approved):
Minor potential scanning bug, please review, looks good otherwise!

ts/packages/benchmarks/src/translationBench/public_datasets/DroidCall/eval/droidCallRawActions.ts
datduyng: Addressed in a5e0714c5e467da7fa90c15ffc0300ee6432ddd4: raw-response matching now compares parameters for chosen actions when rawChosenActions is absent, so a newer same-name attempt with different inputs is rejected. Added coverage for the Ada-versus-Grace case.

PR #2944, opened 2026-08-27
Bind local servers to loopback and gate browser-driven requests

TalZaccai:
## Problem

TypeAgent's servers were reachable from the network. `new WebSocketServer({ port })`
and `server.listen(port)` bind `0.0.0.0` when no host is given, so every one of these
listened on all interfaces even though logs and comments claimed localhost. None of
them authenticate, so any host that could reach the port could drive them:

- **agent server** - `shutdown()`, `joinConversation()`, `getUserIdentity()`. Joining a
  conversation hands back a dispatcher channel, which reaches the file system, mail,
  calendar and browser agents with the local user's permissions.
- **api server** - the `/action/` endpoint dispatches any action from an unauthenticated
  request, including `sendEmail` from the signed-in account.
- **markdown view** - file read/write plus `/agent/execute`. (Fixed separately by #2938;
  this branch no longer touches it.)
- **browser agent bridge, code agent bridge, browser views, montage view** - agent control
  surfaces.
- **shell chat view** - pushed the full conversation to every client that connected.

The api server additionally served static files with `Access-Control-Allow-Origin: *`,
letting any site on the web read them, and ignored the configured `port` entirely.

Two vectors survive a loopback bind, so binding alone was not enough: a page the user
visits can reach `http://localhost:<port>`, and a DNS rebinding attacker can make a
public hostname resolve to loopback.

## Change

Loopback by default, everywhere, with an explicit opt-in where network exposure is a
real deployment need.

- `createWebSocketChannelServer` now binds `127.0.0.1` when given a bare `port` and always
  installs an origin gate. This covers the agent server and the shell discovery server.
- The api server binds loopback unless `TYPEAGENT_API_HOST` (or `host` in config) says
  otherwise, and now honors the configured port. The Dockerfile sets that variable, so
  hosted deployments are unaffected.
- `/action/` and the dispatcher WebSocket refuse requests a web page could have caused,
  checked through `Origin` and `Sec-Fetch-Site`. `TYPEAGENT_API_ALLOWED_ORIGINS` names the
  origins a hosted deployment serves its UI from. An explicit allowlist is used rather than
  comparing `Origin` to `Host`, because a rebinding attacker's request has those two agree.
- Static files echo the request origin only when it passes that same allowlist, and set
  `Vary: Origin`. Same-origin clients need no header at all.
- The remaining listeners bind loopback: montage, browser views, both agent
  bridges, the shell chat view (opt-in via `TYPEAGENT_SHELL_CHAT_SERVER_HOST`), the
  validation demo dashboards, and the `cacheRESTEndpoint` example. The markdown view is
  not included here — #2938 landed the
  same loopback bind for it while this branch was in review, so this branch rebases onto
  that fix rather than duplicating it.
- The onboarding scaffolder templates generate loopback binds, so new agents are not born
  exposed.
- The agent server's `--host` / `AGENT_SERVER_HOST` override trims its value and treats a
  blank one as absent. `--host ""` previously bound every interface, because `??` falls
  through only on `null`/`undefined` and `listen(port, "")` means "all interfaces" — the
  opposite of the default this change establishes.
- `WebSocketChannelServerOptions` no longer accepts `verifyClient`. The factory builds its
  own from `isOriginAllowed`, so one passed in was silently discarded along with the
  caller's access check; it is now a compile error. No existing caller passes one.

Origin gates refuse `Origin: null`, the opaque value a sandboxed iframe sends. A missing
`Origin` stays allowed for non-browser callers such as IoT devices and curl, where the bind
address is the control.

The Visual Studio chat panel is a WebView2 whose content is mapped to a virtual host, so it
sends `Origin: https://typeagent.local` rather than a loopback origin. That origin is
allowed by exact match, alongside the browser-extension schemes. It is deliberately not a
prefix match, which would also accept `https://typeagent.local.evil.example`.

## Known gap

`/action/` accepts GET so simple clients can submit an action as a URL. Browsers older than
roughly Firefox 90 and Safari 16.4 send neither `Origin` nor `Sec-Fetch-Site` on an image
load, and a request carrying neither header is indistinguishable from a non-browser client.
On those browsers a visited page can still reach `/action/` over loopback. Current browsers
label the request `cross-site` and it is refused. This is documented in the api README.

## Validation

New tests cover host resolution, the request gate, the origin allowlist and the CORS
reflection: 35 in the api package (was 28), 13 for the channel server (was 4), and 8 for
agent-server host resolution, including the blank-`--host` case. All three suites pass on
the rebased branch, as does the `agentServer/server` build, which exercises the new
`@typeagent/websocket-utils/loopback` dependency edge.

The three VS Code extensions (`coda`, `vscode-chat`, `vscode-shell`) and the Visual Studio
panel were checked as clients rather than by reading the code. The extensions all run in
the Node extension host, send no `Origin`, and default to `ws://localhost:8999`; driving
the real `connectAgentServer()` against a hardened server connects and lists conversations.
Because the bind is IPv4-only while those clients dial a hostname, `localhost` resolution
was tested explicitly, including with IPv6 forced first — `autoSelectFamily` falls back and
the connection succeeds.

The Visual Studio panel was the one client this broke, and the tests above are what caught
it. Its WebView2 origin now round-trips, its look-alike domain is refused, and both cases
are covered by unit tests.

Binding was confirmed at runtime with `netstat` for the api server, channel server, both
agent bridges, the browser view server, the chat server and the `cacheRESTEndpoint`
example, each of which reported
`127.0.0.1` where it previously reported `0.0.0.0`. Cross-site `/action/` returns 403 and
same-origin returns 200. Static files return a CORS header only for allowlisted origins.

The Electron shell was run in both connect and standalone mode; all 10 listeners bound
loopback and none were reachable from an off-box peer (a WSL2 VM on a separate NAT
network). `@help` round-trips end to end.

The agent-server origin gate was checked against a live server on a confirmed
`127.0.0.1`-only bind:

| Origin | Result |
|---|---|
| *(none)* — native clients (`coda`, `vscode-chat`, CLI) | 101 |
| `http://localhost:8999`, `http://127.0.0.1:5173` | 101 |
| `chrome-extension://…`, `moz-extension://…` | 101 |
| LAN page, `https://evil.example.com`, `null` | 403 |

The cross-origin case was also confirmed in a real browser rather than with a synthetic
header. The same page served from a LAN origin could not open a WebSocket to the agent
server, while the identical page served from loopback could — the loopback case acting as a
positive control so that a blanket failure could not be mistaken for a successful block.
That is the DNS-rebinding vector from the report.

Conversation lifecycle (`listConversations` → `createConversation` → `joinConversation` →
`leaveConversation` → `deleteConversation`) was exercised over the real RPC to confirm
legitimate loopback clients are unaffected.

Scaffolder output was verified end to end: templates were generated through the real
`loadTemplate()`, compiled with `tsc`, and the resulting bridge started and reported a bound
address of `127.0.0.1`.

Lint counts were compared per file against `origin/main` and are unchanged; the new files
add no violations.

### Not covered

Hardware- and service-dependent paths were not exercised: the Android app, Dev Tunnel,
the Docker image, hosted Azure deployment, and the browser extension end to end. The
Visual Studio panel's origin was verified against a live server and in unit tests, but the
panel itself was not driven, since that needs a Visual Studio install. The loopback bind is
IPv4-only, so a client that hard-prefers `::1` would fail; this was not observed here but
was not exhaustively ruled out across all platforms.

For VS Code Remote, both extensions already default to `localhost` and connect from the
extension host with no `Origin`, so the common configurations are unaffected. The one
behavior change is for a client explicitly pointed at a non-localhost address, which is the
exposure this change closes; `AGENT_SERVER_HOST` / `--host` is the documented opt-in. Note
that inside a dev container VS Code's own port forwarding still works, but Docker `-p`
publishing does not, since it can only forward an externally-bound socket.

### DIFF
--- a/ts/examples/cacheRESTEndpoint/src/index.ts
+++ b/ts/examples/cacheRESTEndpoint/src/index.ts
@@ -23,7 +23,7 @@ type SchemaEntry = {
 const app = express();
 
 // Define the port number
-const port = `10482`;
+const port = 10482;
 
 // The cache file to serve
 const cacheFile = "data/v5_sample.json";
@@ -80,8 +80,10 @@ app.get(`/test`, async (req: Request, res: Response) => {
     res.sendFile(path.resolve("src/test.html"));
 });
 
-// Start the server and listen on the specified port
-app.listen(port, async () => {
+// Start the server and listen on the specified port. Bind loopback-only: the
+// endpoint has no authentication and serves cache contents plus a file read
+// from disk, so an all-interfaces bind would expose both to any LAN peer.
+app.listen(port, "127.0.0.1", async () => {
     console.log(`Server is running at http://localhost:${port}`);
     console.log(`Serving cache file from: ${cacheFile}`);
 
--- a/ts/packages/agentServer/server/src/listenHost.ts
+++ b/ts/packages/agentServer/server/src/listenHost.ts
@@ -1,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { LOOPBACK_HOST } from "@typeagent/websocket-utils/loopback";
+
+/**
+ * Resolve the address the agent server listens on: `--host` wins over
+ * `AGENT_SERVER_HOST`, and both fall back to loopback.
+ *
+ * Values are trimmed, and a blank one is treated as absent. Binding the empty
+ * string publishes the listener on every interface, which is the opposite of
+ * the default this function exists to enforce. An untrimmed value is worse
+ * than useless: it fails to resolve at bind time, yet `isLoopbackHost` trims
+ * before comparing, so it would read as loopback and suppress the
+ * network-exposure warning for a host that never binds at all.
+ */
+export function resolveAgentServerHost(
+    argv: readonly string[],
+    env: NodeJS.ProcessEnv,
+): string {
+    const idx = argv.indexOf("--host");
+    const fromArgv = idx !== -1 ? argv[idx + 1]?.trim() : undefined;
+    const fromEnv = env.AGENT_SERVER_HOST?.trim();
+    return fromArgv || fromEnv || LOOPBACK_HOST;
+}
--- a/ts/packages/agentServer/server/src/server.ts
+++ b/ts/packages/agentServer/server/src/server.ts
@@ -8,7 +8,9 @@ import {
 } from "./conversationManager.js";
 import { createAgentServerConnectionHandler } from "./connectionHandler.js";
 import { startStaleBuildWatcher, isStaleBuild } from "./staleBuild.js";
-import { printConfigDriftBanner } from "./banner.js";
+import { printConfigDriftBanner, printWarningBanner } from "./banner.js";
+import { isLoopbackHost } from "@typeagent/websocket-utils/loopback";
+import { resolveAgentServerHost } from "./listenHost.js";
 import {
     getInstanceDirAsync,
     getTraceIdAsync,
@@ -397,6 +399,27 @@ async function main() {
               ? parseInt(process.env.AGENT_SERVER_PORT, 10)
               : AGENT_SERVER_DEFAULT_PORT;
 
+    // The listener is unauthenticated: anyone who can reach the port can join
+    // a conversation and drive the dispatcher with the local user's
+    // permissions. Bind loopback so only this machine can reach it. Remote
+    // access goes through the dev tunnel, whose host process also connects
+    // over loopback, so tunneling still works. `--host` / AGENT_SERVER_HOST
+    // exists for deployments that must publish the port (a container that
+    // isolates the workspace, for example) and warns when it widens the bind.
+    const host = resolveAgentServerHost(process.argv, process.env);
+    if (!isLoopbackHost(host)) {
+        printWarningBanner(
+            [
+                `Agent server is binding ${host}:${port}, not loopback.`,
+                "The server has no authentication: any host that can reach",
+                "this port can join conversations, read your identity, and",
+                "run agent commands as you. Only do this behind a network",
+                "boundary you control.",
+            ],
+            "agent-server",
+        );
+    }
+
     const idleShutdownIdx = process.argv.indexOf("--idle-timeout");
     const idleShutdownMs =
         idleShutdownIdx !== -1
@@ -519,7 +542,7 @@ async function main() {
             },
         });
 
-    wss = await createWebSocketChannelServer({ port }, connectionHandler);
+    wss = await createWebSocketChannelServer({ port, host }, connectionHandler);
 
     // Register the agent-server's own listen port as a regular
     // allocation under the well-known AGENT_SERVER_DISCOVERY_NAME with
@@ -537,7 +560,11 @@ async function main() {
         SYSTEM_SESSION_CONTEXT_ID,
     );
 
-    console.log(`Agent server started at ws://localhost:${port}`);
+    console.log(
+        isLoopbackHost(host)
+            ? `Agent server started at ws://localhost:${port}`
+            : `Agent server started at ws://${host}:${port}`,
+    );
     writeServerPid(port, process.pid);
     // Publish our own listen URL into the process environment so in-process
     // consumers (e.g. the reasoning loop's subagent manager, which spawns
--- a/ts/packages/agents/onboarding/src/scaffolder/templates/websocketBridgeTemplate.ts
+++ b/ts/packages/agents/onboarding/src/scaffolder/templates/websocketBridgeTemplate.ts
@@ -95,7 +95,14 @@ export class __AgentName__Bridge {
      */
     public static start(port: number = 0): Promise<__AgentName__Bridge> {
         return new Promise((resolve, reject) => {
-            const server = new WebSocketServer({ port });
+            const server = new WebSocketServer({
+                port,
+                // Bind loopback. The bridge carries agent RPC with the local
+                // user's permissions and authenticates nothing, so an
+                // all-interfaces bind would let any LAN peer impersonate the
+                // plugin, intercept requests, and forge responses.
+                host: "__LOOPBACK_HOST__",
+            });
             let settled = false;
             const onError = (e: Error) => {
                 if (settled) return;
--- a/ts/packages/utils/webSocketChannelServer/src/server.ts
+++ b/ts/packages/utils/webSocketChannelServer/src/server.ts
@@ -8,6 +8,11 @@ import {
 import WebSocket, { WebSocketServer } from "ws";
 import registerDebug from "debug";
 import { createPromiseWithResolvers } from "@typeagent/common-utils";
+import {
+    createConfiguredOriginAllowlist,
+    VISUAL_STUDIO_WEBVIEW_ORIGIN,
+} from "@typeagent/websocket-utils/originAllowlist";
+import { LOOPBACK_HOST } from "@typeagent/websocket-utils/loopback";
 import { attachHeartbeat } from "./heartbeat.js";
 
 const debugWss = registerDebug("typeagent:transport:wss");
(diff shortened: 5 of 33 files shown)


ts/packages/utils/webSocketChannelServer/test/server.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+    return new Promise((resolve) => {
+        const ws = new WebSocket(
+            `ws://${host}:${port}`,
+            origin === undefined ? undefined : { origin },
+        );
+        openClients.push(ws);
+        ws.once("open", () => resolve("open"));
+        ws.once("error", () => resolve("refused"));
+    });
+}
```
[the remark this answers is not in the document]
TalZaccai: Declining this one — the hang can't happen with `ws`.

In `ws`'s client, the `unexpected-response` path is:

```js
} else if (!websocket.emit('unexpected-response', req, res)) {
  abortHandshake(websocket, req, new Error(`Unexpected server response: ${res.statusCode}`));
}
```

`EventEmitter.emit()` returns `false` when there are no listeners for the event, so with no `unexpected-response` handler attached the `!` branch always runs, `abortHandshake` fires, and `error` is emitted with `Unexpected server response: 403`. I checked both `ws` versions installed in this repo (7.5.13 and 8.21.3) — identical in each.

That also matches observed behavior: the suite has three refusal cases that assert on the rejection, and all 13 tests pass. If `error` weren't emitted they would time out rather than pass.

Adding an `unexpected-response` listener would actually change behavior here — it suppresses `abortHandshake`, so the socket would be left open and the helper really would hang.

ts/packages/utils/webSocketChannelServer/src/server.ts
```diff
@@ -29,18 +31,36 @@ type WebSocketChannelServer = {
...
+ */
+const defaultIsOriginAllowed = createAgentOriginAllowlist({
+    extensionSchemes: ["chrome-extension://", "moz-extension://"],
+    allowNullOrigin: false,
+});
+
 /**
  * Extra options layered on top of `ws.ServerOptions` for our transport.
  */
 export type WebSocketChannelServerOptions = WebSocket.ServerOptions & {
```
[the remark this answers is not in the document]
TalZaccai: Good catch — fixed in 674b5b957.

The option type is now `Omit<WebSocket.ServerOptions, "verifyClient"> & {...}`, so passing one is a compile error instead of a silently dropped access check, with a doc comment pointing callers at `isOriginAllowed`.

Non-breaking: the only two production callers are `agentServer/server/src/server.ts` and `shell/src/main/discoveryServer.ts`, and neither passes `verifyClient`.

ts/packages/agentServer/server/src/server.ts
```diff
@@ -397,6 +401,32 @@ async function main() {
...
+    // access goes through the dev tunnel, whose host process also connects
+    // over loopback, so tunneling still works. `--host` / AGENT_SERVER_HOST
+    // exists for deployments that must publish the port (a container that
+    // isolates the workspace, for example) and warns when it widens the bind.
+    const hostIdx = process.argv.indexOf("--host");
+    const hostOverride =
+        hostIdx !== -1
+            ? process.argv[hostIdx + 1]
+            : process.env.AGENT_SERVER_HOST?.trim() || undefined;
+    const host = hostOverride ?? LOOPBACK_HOST;
```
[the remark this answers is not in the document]
TalZaccai: Fixed in 674b5b957 — and it was worse than described.

`??` only falls through on `null`/`undefined`, so `--host ""` produced `""`, and `listen(port, "")` binds **every** interface — the exact opposite of the default this PR exists to establish. The trimming gap was a second problem: an untrimmed `" localhost "` fails to resolve at bind time, but `isLoopbackHost` trims before comparing, so it read as loopback and suppressed the network-exposure warning for a host that never binds.

Extracted to `resolveAgentServerHost(argv, env)` in `listenHost.ts`, which trims both sources and uses `||` so a blank value is treated as absent (argv → env → loopback). Covered by 8 unit tests in `test/listenHost.spec.ts`, including the blank-`--host` case and `--host` appearing as the last argument.

ts/examples/cacheRESTEndpoint/src/index.ts
```diff
@@ -23,7 +23,7 @@ type SchemaEntry = {
 const app = express();
 
 // Define the port number
-const port = `10482`;
```
robgruen, line 26: ahhh!!!!

ts/packages/agents/onboarding/src/scaffolder/templates/websocketBridgeTemplate.ts
```diff
@@ -95,7 +95,14 @@ export class __AgentName__Bridge {
...
     public static start(port: number = 0): Promise<__AgentName__Bridge> {
         return new Promise((resolve, reject) => {
-            const server = new WebSocketServer({ port });
+            const server = new WebSocketServer({
+                port,
+                // Bind loopback. The bridge carries agent RPC with the local
+                // user's permissions and authenticates nothing, so an
+                // all-interfaces bind would let any LAN peer impersonate the
+                // plugin, intercept requests, and forge responses.
+                host: "127.0.0.1",
```
robgruen: maybe a constant since we use this in several places in this package

ts/packages/agentServer/server/src/listenHost.ts
```diff
@@ -0,0 +1,25 @@
...
+ * `AGENT_SERVER_HOST`, and both fall back to loopback.
+ *
+ * Values are trimmed, and a blank one is treated as absent. Binding the empty
+ * string publishes the listener on every interface, which is the opposite of
+ * the default this function exists to enforce. An untrimmed value is worse
+ * than useless: it fails to resolve at bind time, yet `isLoopbackHost` trims
+ * before comparing, so it would read as loopback and suppress the
+ * network-exposure warning for a host that never binds at all.
+ */
+export function resolveAgentServerHost(
```
robgruen, line 17: can we specify more than one?  On some machines you might want to do loopback & public ip or there might be multiple NICs or IPv4 & IPv6

robgruen (approved):

ts/examples/cacheRESTEndpoint/src/index.ts
TalZaccai: Ha — yes. `` const port = `10482` `` was a template literal holding a number, so `app.listen(port, ...)` was taking the string overload.

That's also why the loopback fix needed a second change: Express has no `(string, string, callback)` overload, so adding the host argument didn't compile until `port` became an actual number.

ts/packages/agents/onboarding/src/scaffolder/templates/websocketBridgeTemplate.ts
TalZaccai: Done in 797e2ca25 — three occurrences, in `websocketBridgeTemplate.ts`, `websocketBridgeHandler.ts` and `viewUiHandler.ts`.

One wrinkle worth flagging, since it shaped the fix: these files are templates, read as raw text by `loadTemplate()` and emitted verbatim into the scaffolded agent. So an ordinary exported constant wouldn't have worked — importing `LOOPBACK_HOST` from `@typeagent/websocket-utils` would have been emitted into the generated package, which doesn't depend on it, and the generated code wouldn't resolve the import.

Instead I added it as a template variable. `templateLoader.ts` now carries a `DEFAULT_TEMPLATE_VARS` map with `LOOPBACK_HOST: "127.0.0.1"`, merged under caller-supplied vars so a call site can still override, and the templates say `host: "__LOOPBACK_HOST__"`. One definition in the package, and the emitted code keeps a plain literal with no new dependency.

I kept the constant local rather than importing the shared `LOOPBACK_HOST`, since onboarding doesn't currently depend on `@typeagent/websocket-utils` and adding a dependency for one string would churn the lockfile.

Verified the rendered output is unchanged: all three templates still emit `"127.0.0.1"` with no leftover placeholder, and the package builds clean.

ts/packages/agentServer/server/src/listenHost.ts
TalZaccai: Not today, and not without real work — but the common cases are already covered.

Node binds one address per server: `listen()` takes a single `host`, and `ws` passes it straight through. Binding a *set* of addresses means running N server instances and managing their combined lifecycle (readiness, partial-bind failure, shared shutdown, and what `address()` reports). That's a feature rather than a hardening change, so I'd rather not fold it into this PR.

What you can get right now, all through the existing `--host` / `AGENT_SERVER_HOST` opt-in:

- one specific address, e.g. `--host 192.168.1.50`
- every IPv4 interface, `--host 0.0.0.0`
- **every interface on both families, `--host ::`** — I checked this rather than assuming: `host: "::"` binds dual-stack, and a client on `::1` *and* a client on `127.0.0.1` both connect to that one listener.

So "loopback and public IP" and "IPv4 and IPv6" both work today via `::`. What genuinely isn't expressible is an arbitrary *subset* — loopback plus one NIC while excluding the others — since there's no single address that denotes it.

Worth noting the default is deliberately `127.0.0.1` and not `::`: IPv4 loopback is what every in-repo client resolves to, and `::` would put the port on every interface, which is the exposure this PR closes.

One real limitation this leaves, already called out under "Not covered" in the description: the default is IPv4-only, so a client that explicitly dials `[::1]` fails. Clients using the hostname `localhost` are fine — Node's `autoSelectFamily` falls back to IPv4, which I verified with `--dns-result-order=ipv6first`.

Happy to file a follow-up issue for multi-address binding if you want it tracked, and equally happy to reopen this if you'd rather it land here.

robgruen:
OH, I didn't notice.  If it's a template just leave it...may get unexpected behavior if the referencing variable changes.

PR #2955, opened 2026-09-01
[Telemetry] Add @log open to quickly open traces

GeorgeNgMsft:
Add @log open <traceId|last> to quickly open traces in local Grafana view (opens in new window + chat link)

image

image

### DIFF
--- a/ts/packages/dispatcher/dispatcher/src/context/commandHandlerContext.ts
+++ b/ts/packages/dispatcher/dispatcher/src/context/commandHandlerContext.ts
@@ -302,6 +302,14 @@ export type PendingTopicalRoute = {
 };
 
 // Command Handler Context definition.
+export type SessionTrace = {
+    traceId: string;
+    requestId: string;
+    kind: "command" | "request";
+    isTraceOpen: boolean;
+    completedAt: number;
+};
+
 export type CommandHandlerContext = {
     readonly agents: AppAgentManager;
     readonly portRegistrar: IPortRegistrar;
@@ -407,15 +415,10 @@ export type CommandHandlerContext = {
      */
     readonly traceId: string | undefined;
     /**
-     * Canonical OpenTelemetry root trace id of the previous completed user
-     * request, used to resolve `@log open last`.
+     * Canonical OpenTelemetry root traces completed in the current session.
+     * Kept in completion order for trace inspection commands.
      */
-    lastCommandResultTraceId?: string | undefined;
-    /**
-     * Set while processing a user request. Diagnostic trace-open actions clear
-     * it so they do not replace the trace they are inspecting.
-     */
-    rememberCurrentRequestTrace: boolean;
+    readonly sessionTraceHistory: SessionTrace[];
     readonly telemetryOptions: {
         readonly joinActiveTrace: boolean;
     };
@@ -1340,8 +1343,7 @@ export async function initializeCommandHandlerContext(
             logger,
             activationId,
             traceId,
-            lastCommandResultTraceId: undefined,
-            rememberCurrentRequestTrace: false,
+            sessionTraceHistory: [],
             telemetryOptions: {
                 joinActiveTrace: options?.telemetry?.joinActiveTrace ?? false,
             },
@@ -1995,6 +1997,7 @@ export async function setSessionOnCommandHandlerContext(
     session: Session,
 ) {
     context.session = session;
+    context.sessionTraceHistory.length = 0;
     await context.agents.close();
 
     await initializeMemory(context, session.getSessionDirPath());
--- a/ts/packages/dispatcher/dispatcher/src/context/commandHandlerContext.ts
+++ b/ts/packages/dispatcher/dispatcher/src/context/commandHandlerContext.ts
@@ -406,6 +406,16 @@ export type CommandHandlerContext = {
      * `typeagent.trace.id` so existing logs can still be joined.
      */
     readonly traceId: string | undefined;
+    /**
+     * Canonical OpenTelemetry root trace id of the previous completed user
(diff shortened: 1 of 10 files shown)


GeorgeNgMsft:
Minor CICD fix here and doc changes waiting on other PRs to complete

datduyng (approved):
@GeorgeNgMsft, my suggestion

It's good practice to persist all trace id some how - either through storage or in memory (for the active session)



Example of claude jsonl

Image

The uuid is the server id

ts/packages/dispatcher/dispatcher/src/context/commandHandlerContext.ts
```diff
@@ -406,6 +406,16 @@ export type CommandHandlerContext = {
      * `typeagent.trace.id` so existing logs can still be joined.
      */
     readonly traceId: string | undefined;
+    /**
+     * Canonical OpenTelemetry root trace id of the previous completed user
+     * request, used to resolve `@log open last`.
+     */
+    lastCommandResultTraceId?: string | undefined;
```
datduyng: code looks fine. 
 
From an interface perspective, I don't think command handler should keep only the last traceId.
 
We should look if we can hold all traces and get the last one

ts/packages/dispatcher/dispatcher/src/context/commandHandlerContext.ts
```diff
@@ -406,6 +406,16 @@ export type CommandHandlerContext = {
...
+    /**
+     * Set while processing a user request. Diagnostic trace-open actions clear
+     * it so they do not replace the trace they are inspecting.
+     */
+    rememberCurrentRequestTrace: boolean;
```
datduyng: We should remove this - it should just persist by default - since trace id is essential

ts/packages/dispatcher/dispatcher/src/context/commandHandlerContext.ts
GeorgeNgMsft: Updated this to retain ordered metadata for every valid root trace in the active session. `@log open last` now derives the newest non-inspection request from that history, leaving the collection available for a follow-up trace-listing feature.

ts/packages/dispatcher/dispatcher/src/context/commandHandlerContext.ts
GeorgeNgMsft: Removed the opt-in persistence flag. Every valid root trace is now appended by default; trace-open requests are retained and classified so they can be excluded only when resolving `last`.

ts/packages/dispatcher/dispatcher/src/context/system/handlers/logCommandHandler.ts
```diff
@@ -194,7 +202,316 @@ function getConstructorName(value: unknown): string {
     return ctor?.name ?? "unknown";
 }
 
-export function getLogCommandHandlers(): CommandHandlerTable {
+export const LOCAL_GRAFANA_BASE_URL = "http://127.0.0.1:24319";
+
+const LOCAL_GRAFANA_HEALTH_URL = `${LOCAL_GRAFANA_BASE_URL}/api/health`;
+const LOCAL_GRAFANA_HEALTH_TIMEOUT_MS = 1500;
+const LOCAL_TEMPO_TRACE_WAIT_ATTEMPTS = 20;
+const LOCAL_TEMPO_TRACE_WAIT_INTERVAL_MS = 500;
```
jebrans: worst case it will wait for 40 seconds attempt * (timeout + wait time) , is that ok?

ts/packages/dispatcher/dispatcher/src/context/system/handlers/logCommandHandler.ts
```diff
@@ -194,7 +202,316 @@ function getConstructorName(value: unknown): string {
...
+            await waitForRetry(dependencies, abortSignal);
+        }
+    }
+    return backendUnavailable ? "unavailable" : "not-found";
+}
+
+function combineAbortSignals(
+    abortSignal: AbortSignal | undefined,
+): AbortSignal {
+    const timeoutSignal = AbortSignal.timeout(LOCAL_GRAFANA_HEALTH_TIMEOUT_MS);
```
jebrans: We use same timeout for fetch. If tempo query takes longer than 1.5 sec we will abort.

jebrans (approved):

ts/packages/dispatcher/dispatcher/src/context/system/handlers/logCommandHandler.ts
GeorgeNgMsft: Good call out, 40s doesn't make sense in practice

ts/packages/dispatcher/dispatcher/src/context/system/handlers/logCommandHandler.ts
GeorgeNgMsft: Separated the timeouts: Grafana's lightweight health probe keeps its 1.5-second timeout, while the Tempo fetch uses the shared 5-second overall lookup signal. A slow valid Tempo query therefore gets the remaining lookup budget instead of being aborted at 1.5 seconds.

ts/packages/dispatcher/dispatcher/src/context/system/handlers/logCommandHandler.ts
GeorgeNgMsft: Updated this to use one 5-second overall Tempo lookup budget rather than multiplying a per-fetch timeout by 20 attempts. Fast 404s are retried every 500 ms, but the combined query-and-wait path cannot exceed that shared budget.

PR #2963, opened 2026-09-02
Trace client agent cleanup failures and pin the group's method set

jebrans:
Two follow-ups to #2914, both on the client-hosted agent path. Independent of each other; either can be dropped.

### 1. Disconnect cleanup was silent

`connectionHandler.ts` swallowed every failure from the disconnect path:

- `removeClientAgent` failing is exactly how a client agent leaks onto the shared dispatcher.
- `leaveConversation` failing keeps a dispatcher alive and its idle timer from ever starting.

Both then surface much later as unrelated-looking failures, with nothing pointing back at the cause. There is no one to report them to -- the socket is already gone -- so they are traced on `agent-server:connection:error`.

**Not retried, on purpose.** Removal is idempotent and ownership-checked (`removeClientAgentInstance` returns false when the instance is absent or belongs to another connection), so a second attempt could only race a reconnect that has legitimately reclaimed the instance.

### 2. A device could advertise methods it does not implement

`createMux` builds its method set from `Object.keys(template)` on whichever proxy created the group, and `getManifestKey` hashes **schema text only**. Two app builds can therefore share a schema and still implement a different `agentInterface`. The narrower device joined and silently appeared to support the richer set; the call only failed once someone actually made it.

This compares the normalized `agentInterface` at join, next to the existing schema check:

- Covers a **replacement** too, not just a new instance -- replacing in place keeps the mux built from the original proxy.
- Order-insensitive and de-duplicated, so key order cannot cause a false mismatch (same trap `getManifestKey` already avoids for Android's `org.json.JSONObject`).
- `agentInterface` is **required**, so there is no client that skips the check. `registerClientAgent` declares it required and `createAgentRpcClient` dereferences it to build the proxy, so a registration that reaches the registry always carries one.
- A **lone device** that upgrades its app (same schema, method set grows or shrinks) adopts the new set instead of being rejected. Nobody else is sharing the name, so there is no other device to conflict with -- and nothing for the user to disconnect. The mux is updated in place, because the dispatcher keeps the object `addDynamicAgent` handed it and checks optional methods on it at call time.

Deliberately kept out of `getManifestKey`: a separate key lets the error say which of the two things actually differs.

Not reachable in production today -- the Android agent's `agentInterface` is `["executeAction"]` only -- but it becomes live the moment that list grows.

### 3. The Android client could grow that list without noticing

`AndroidDeviceAgent` hardcoded `agentInterface` as `["executeAction"]`, and `handleAndroidDeviceInvoke` separately hardcoded the same string as its dispatch guard. Nothing tied the two together, and no CI job builds `android/`, so adding a method to one and not the other would go unnoticed -- producing exactly the mismatch above, in the direction that fails at the call.

Both now come from `AndroidDeviceAgent.SUPPORTED_METHODS`. The unit test asserts the declared array against that list rather than a literal, and pins that an unimplemented method is not claimed, so widening the list without adding dispatch fails the test.

### Tests

8 new cases in `clientAgentRegistry.spec.ts`. `makeDevice` builds a proxy carrying exactly the methods it declares, so the cases exercise the actual misroute rather than only the string comparison: a second device with fewer methods is rejected (and the mux keeps the method it advertises); one with extra methods is rejected; the same set reordered is accepted; the empty set is compared like any other; a plain reconnect keeps its slot; a lone device gaining a method has it routed through the same mux; a lone device losing one has it taken off the mux; and a reconnecting device cannot change a *shared* group's set.

**Mutation-checked.** Disabling the rejection fails only the four rejection cases; disabling the in-place mux update fails only the two cases that assert on the mux. Acceptance cases pass throughout.

### Verification

- `pnpm --filter agent-server build` -- clean.
- `pnpm --filter agent-server test` -- **119 passed** (111 before).
- `npm run code-lint -- --ratchet --base origin/main` -- `violations base 6 -> head 6`, OK.
- `pnpm run prettier:changed` -- clean.
- `gradlew assembleDebug testDebugUnitTest` (mobile-2) -- **171 passed**; nothing in CI covers `android/`.

### DIFF
--- a/ts/packages/agentServer/server/src/clientAgentRegistry.ts
+++ b/ts/packages/agentServer/server/src/clientAgentRegistry.ts
@@ -37,6 +37,11 @@ export type ClientAgentGroup = {
     /** Hash of the schema source; instances must agree on it. See {@link getManifestKey}. */
     manifestKey: string;
     /**
+     * Normalized `agentInterface` the group was created with, or undefined when
+     * the creator did not declare one. See {@link getAgentInterfaceKey}.
+     */
+    agentInterfaceKey: string | undefined;
+    /**
      * Whether the client that created this group opted in to sharing the name.
      * Off means a second, different client is rejected exactly as it was before
      * groups existed, so a client that assumes it is the only host of the name
@@ -53,6 +58,11 @@ export type ClientAgentRegistration = {
     connectionId: string;
     appAgent: AppAgent;
     manifest: AppAgentManifest;
+    /**
+     * Methods the client implements. Optional so a client that does not send
+     * one keeps working; when present it must match the group's.
+     */
+    agentInterface?: readonly string[] | undefined;
     /** See {@link ClientAgentGroup.multiInstance}. Only read on the first registration. */
     multiInstance?: boolean;
 };
@@ -133,6 +143,28 @@ export function getManifestKey(manifest: AppAgentManifest): string {
 
 export function schemaMismatchMessage(name: string): string {
     return `Client agent '${name}' is already registered on this conversation with a different schema version. Update the app to the same version as the other device(s), or disconnect them first.`;
+}
+
+/**
+ * Normalized `agentInterface`, or undefined when the client did not declare
+ * one.
+ *
+ * The mux is built once, from the first instance's proxy, and
+ * {@link getManifestKey} only covers schema text -- two app versions can share
+ * a schema and still implement different methods. Without this, a device with a
+ * narrower interface joins a group created by a richer one and silently appears
+ * to support methods it does not; the call only fails once someone makes it.
+ */
+export function getAgentInterfaceKey(
+    agentInterface: readonly string[] | undefined,
+): string | undefined {
+    return agentInterface === undefined
+        ? undefined
+        : [...new Set(agentInterface)].sort().join("\u0000");
+}
+
+export function interfaceMismatchMessage(name: string): string {
+    return `Client agent '${name}' is already registered on this conversation by a device that implements a different set of methods. Update the app to the same version as the other device(s), or disconnect them first.`;
 }
 
 /**
@@ -530,6 +562,7 @@ export function createClientAgentGroup(
         name,
         manifest: registration.manifest,
         manifestKey: getManifestKey(registration.manifest),
+        agentInterfaceKey: getAgentInterfaceKey(registration.agentInterface),
         multiInstance: registration.multiInstance === true,
         instances: new Map([[instance.instanceId, instance]]),
         mux: undefined as unknown as AppAgent,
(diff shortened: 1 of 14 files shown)


ts/packages/agentServer/server/src/clientAgentRegistry.ts
```diff
@@ -561,6 +636,27 @@ export async function joinClientAgentGroup(
...
+        // Unless this device is the whole group. A lone device that upgrades
+        // its app keeps its schema but can gain or lose methods, and rejecting
+        // that would break a reconnect that used to work - with a message
+        // telling the user to disconnect devices that do not exist. Nobody else
+        // is sharing the name, so adopt the new set and bring the mux with it.
+        if (!replacesSoleInstance(group, registration)) {
+            throw new Error(interfaceMismatchMessage(group.name));
+        }
+        group.agentInterfaceKey = agentInterfaceKey;
+        rebuildMux(group, registration.appAgent);
```
GeorgeNgMsft: It maybe better to reload the agent instead of rebuilding the mux because a new device upgrade may add new methods, but this isn't running initialization right now. This may miss dependent state changes despite loading new capabilities (which would then potentially fail).

ts/packages/agentServer/server/test/clientAgentRegistry.spec.ts
```diff
@@ -257,6 +292,205 @@ describe("clientAgentRegistry registration", () => {
...
+    test("a lone device that upgrades its app changes the group's method set", async () => {
+        const registry = createClientAgentRegistry();
+        const host = makeHost();
+
+        await register(registry, host, {
+            instanceId: "a",
+            connectionId: "conn-a",
+            appAgent: makeDevice(["executeAction"]).appAgent,
+        });
+        expect(getMux(registry).getDynamicDisplay).toBeUndefined();
```
GeorgeNgMsft: You may want to hold a reference to the original mux here
`const dispatcherMux = host.registered.get(AGENT_NAME)!;`

and check that same object later to ensure that it's properly mutated. 
Currently `getMux(registry)` below makes it possible for an incorrect implementation to change the mux that is being returned. 

This ensures that the changes are applied to the original dispatcher mux.

GeorgeNgMsft (approved):
Lgtm! A couple minor comments for consideration

ts/packages/agentServer/server/src/clientAgentRegistry.ts
jebrans: Addressed in e7633d368. Re-registration now uses a unique RPC channel so the old endpoint remains live while the replacement is validated. A sole-device method-set change replaces the dynamic agent through full dispatcher lifecycle setup, checks setup failures, and restores the previous registration on failure.

ts/packages/agentServer/server/test/clientAgentRegistry.spec.ts
jebrans: Addressed in e7633d368. The tests now retain the dispatcher-owned mux reference from host.registered and assert the intended identity explicitly: unchanged-interface replacement keeps that mux, while a method-set reload installs a new dispatcher mux and leaves the old reference untouched.

PR #2964, opened 2026-09-02
Add more androidDevice actions to the mobile-2 sample

jebrans:
### What this changes

- Adds `composeEmail`, `shareText`, `openSettings`, `createCalendarEvent` and
  `playMusicFromSearch` to `androidDeviceSchema.ts` and wires each one through
  `AndroidDeviceAgent`, `WebSocketManager`, `ChatViewModel` and `MainActivity`.
- Uses `ACTION_SENDTO` with a bare `mailto:` URI for `composeEmail`, so only
  mail apps resolve rather than every share target, and carries recipients as
  `EXTRA_EMAIL`/`EXTRA_CC`/`EXTRA_BCC` arrays instead of splicing them into the
  URI. A single unusable address fails the whole action, because a draft
  addressed to fewer people than the user asked for would look like success.
- Resolves the inner `ACTION_SEND` intent before wrapping it in
  `Intent.createChooser` for `shareText`. The chooser is a system activity that
  always resolves, so an unchecked wrap would report success on a device with
  no text handler.
- Restricts `openSettings` to a closed `AndroidSettingsScreen` enum mapped
  exhaustively to `Settings.ACTION_*` constants. The model never supplies a raw
  intent action string, `appInfo` is pinned to this app's own package, and the
  screens only display settings rather than toggling them.
- Adds `createCalendarEvent` as `ACTION_INSERT` on `CalendarContract.Events`,
  which opens the calendar app's pre-filled new-event editor and so needs no
  `WRITE_CALENDAR` permission. ISO-8601 times are resolved without `java.time`,
  since the module targets `minSdk 24` with no core-library desugaring.
- Builds calendar offset zones from a raw integer offset via `SimpleTimeZone`
  rather than by formatting a `GMT+hh:mm` string. `String.format` is
  locale-sensitive, and under a locale with non-Latin digits
  `TimeZone.getTimeZone` cannot read the result and silently returns GMT,
  turning `+05:30` into a five-and-a-half-hour shift with no error.
- Anchors all-day events at UTC midnight as `CalendarContract` requires, so a
  user in UTC+10 does not see them land a day early.
- Fails rather than guessing on an unknown music `focus`, an unknown settings
  screen, an all-day event that also carries a time of day, and an event
  spanning more than 366 days.
- Adds a `<queries>` entry for every new implicit intent. Without one, Android
  11+ package visibility makes `resolveActivity` return null and the action
  falsely reports that no app is available.
- Extends `AndroidDeviceSchemaAssetTest` to check that the settings-screen and
  music-focus unions in the schema match the Kotlin enums, so the two cannot
  drift apart silently.

Every action reaches the OS through the existing `launchExternalIntent` funnel,
keeping the current `resolveActivity` check, foreground-lifecycle check and
exception handling, and completing its callback exactly once. The ten existing
Tier-1 actions are unchanged.

### Verification

Unit tests and the debug build completed successfully:

- `gradlew testDebugUnitTest` — 234 tests, 0 failures (previously 179).
- `gradlew assembleDebug` — successful.
- `gradlew lintDebug` — clean; the only findings are pre-existing
  informational items unrelated to this change.

Review confirmed:

- A GPT-5.3-Codex review reported no Critical or High findings. Its Medium
  finding (locale-sensitive offset formatting) and Low finding (fractional
  seconds rejected) are both fixed here, each with a regression test. The
  locale defect was reproduced on the JVM before and after the fix: `+05:30`
  resolved to offset `0` instead of `19800000`.
- Hours-only ISO offsets such as `+05` are now accepted, with a test.
- No new permissions are declared in `AndroidManifest.xml`.


android/samples/mobile-2/app/src/main/java/com/example/typeagentchat/ShareTextActionParser.kt
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * The text itself is required and never defaulted. There is no sensible "share
+ * whatever was on screen" behaviour to fall back to, and inventing one would
+ * risk sharing conversation content the user never pointed at.
+ */
+internal fun parseShareTextActionPayload(data: Any?): ShareTextAction? {
+    val payload = data as? JSONObject ?: return null
+    // Newlines survive here, unlike in the URI-bound actions: shared text is
+    // carried as an extra, and a multi-line note is a normal thing to share.
+    val text = payload.optActionString("text").trim().take(MAX_SHARE_TEXT_CHARS).trim()
```
GeorgeNgMsft: If the text is larger than the limit, it would be a good idea to surface some sort of alert or exception as this is a silent truncation.

android/samples/mobile-2/app/src/main/java/com/example/typeagentchat/ComposeEmailActionParser.kt
```diff
@@ -0,0 +1,135 @@
...
+ * than they asked for, and the model would never learn it got the address wrong.
+ */
+internal fun parseComposeEmailActionPayload(data: Any?): ComposeEmailAction? {
+    val payload = data as? JSONObject ?: return null
+
+    val to = payload.parseEmailAddressList("to") ?: return null
+    val cc = payload.parseEmailAddressList("cc") ?: return null
+    val bcc = payload.parseEmailAddressList("bcc") ?: return null
+    val subject = payload.sanitizedActionText("subject")
+    val body = payload.sanitizedActionText("body", MAX_EMAIL_BODY_CHARS)
```
GeorgeNgMsft: This might destroy the line breaks and whitespace, you should double check if it does and ensure that i preserves the spacing as appropriate

GeorgeNgMsft (approved):
Left a couple minor considerations, looks good though!

android/samples/mobile-2/app/src/main/java/com/example/typeagentchat/ShareTextActionParser.kt
jebrans: Fixed

PR #2966, opened 2026-09-03
[Markdown] create documents in workspace

GeorgeNgMsft:
## Summary

This PR fixes document creation and separates open/create semantics

Layer 1 (bottom) of the replacement stack for #2940.

This layer adds only core workspace-safe Markdown document creation:

- propagates the host-authorized working directory through action context and agent RPC
- creates Markdown files under that workspace, including nested relative paths
- preserves requested initial content
- rejects traversal, absolute/drive-qualified, out-of-workspace, and symlink-escape paths
- avoids constructing the Markdown model for create/open actions
- passes the CLI request working directory into dispatch

## Deferred to upper layers

- document update/edit semantics
- revision and binding-token conflict checks
- streaming update persistence
- browser-view persistence and synchronization
- reopening editor bindings
- loopback view-server changes
- the unrelated `aiclient` runtimeConfig/OpenAI change

## Validation

- `pnpm run prettier:changed:fix` / `pnpm run prettier:changed` (pass)
- lint ratchet (pass)
- complexity ratchet (pass)
- circular-dependency ratchet (pass)
- debt/skipped-test gate (pass)
- direct path-policy smoke covering nested creation, traversal rejection, canonical root resolution, and junction escape rejection (pass)
- affected TypeScript build and Jest suites were attempted but environment-blocked: worktree setup could not fetch locked public packages `baseline-browser-mapping@2.11.19` and `uuid@14.0.2` because Azure Artifacts returned 401, leaving workspace package links incomplete; the resulting compiler errors were missing-module diagnostics

PR #2940, opened 2026-08-27
fix(markdown): restore document creation and updates

GeorgeNgMsft:
## Summary

- restore Markdown document creation and updates without requiring a connected browser
- preserve initial content for compound create-and-fill requests and avoid overwriting existing non-empty files
- save documents under the host-authorized workspace root, with safe relative subdirectories and traversal/symlink/junction protections
- propagate the request working directory through CLI, dispatcher `ActionContext`, and out-of-process agent RPC
- return the real filesystem path and a loopback editor link for every created/opened document
- unify browser and headless Markdown operation semantics, including formatting, serializer-based persistence, revisions, binding identities, restart recovery, and multi-tab primary handoff
- remove the unsafe unused `/file/load` rebinding path and evict stale collaboration rooms
- preserve typed Azure model configuration while deferring LLM initialization until an update actually needs it

## Validation

- affected Markdown, RPC, dispatcher, CLI, and server builds pass
- 103 Markdown tests pass across 9 suites
- 38 agent RPC tests pass across 3 suites
- focused dispatcher `ActionContext` tests pass
- complexity and lint ratchets pass; changed files are Prettier-clean
- real CLI compound request executed `markdown.createDocument`, wrote exact content at the repository root, appeared in `git status`, and returned an editor URL that responded successfully
- completed two principal-engineer adversarial review rounds plus a narrow closure review; all high-confidence findings were addressed

PR #2968, opened 2026-09-03
fix(markdown): persist conflict-safe updates

GeorgeNgMsft:
## Stack

Layer 3 of the replacement stack for #2940.

- Base: `main` (#2966 and #2967 are merged)
- Parent layer: #2967 (merged)

## Included

- Persists headless Markdown updates directly to the authorized workspace file.
- Makes streaming completion idempotent by recognizing the already-produced revision instead of applying operations twice.
- Carries immutable binding identity and SHA-256 revision checks across read/apply.
- Rejects stale tokens, rebound root/path/file identities, revision conflicts, and roots replaced by symlinks or junctions.
- Adds request IDs to concurrent document read/apply IPC and correlates responses.
- Keeps the view service server-authoritative for update persistence even when no browser/SSE client is connected.

## Deferred to browser layer

- Browser `document-manager.ts` rewrite and serializer/autosave behavior.
- Primary/secondary SSE lifecycle, browser synchronization/promotion UX, and reopening the same file while preserving browser binding state.
- Broader browser snapshot adoption and loopback binding behavior.

## Validation

Historical results from before the rebase (not rerun as full build/Jest validation):

- `@typeagent/markdown-agent` TypeScript and frontend build
- Focused Jest suites: 44 tests passed across Markdown persistence, operation engine, path policy, creation path policy, and action handler coverage
- Prettier changed-file check
- Git diff check
- Repo lint ratchet

After rebasing onto `f3e4308ea2cd722a2905eceac3f2a1e61d19a420`: 23 focused Node integration smoke checks passed using the actual handler, persistence module, and operation engine with simulated translator/view IPC; TypeScript 5.4.5 syntax parsing and Git diff checks passed. Full build/Jest, semantic typechecking, and repository ratchets were not rerun: this worktree has no installed dependencies (`fluid-build` missing). Pinned Prettier 3.5.3 was unavailable offline.

Complete layer diff against the new main base: 992 insertions, 532 deletions (1,524 changed lines), across 6 files.

### DIFF
--- a/ts/packages/agents/markdown/src/view/route/service.ts
+++ b/ts/packages/agents/markdown/src/view/route/service.ts
@@ -22,9 +22,7 @@ import { randomUUID } from "node:crypto";
 import { isAllowedViewOrigin } from "./originAllowlist.js";
 import { resolvePathWithinRoot } from "./pathPolicy.js";
 import {
-    isCanonicalDirectory,
     normalizeRelativeDocumentPath,
-    resolveExistingFileWithinRoot,
     resolveRealDirectory,
     resolveWritableFileWithinRoot,
 } from "../../agent/pathPolicy.js";
@@ -262,8 +260,16 @@ const INITIAL_ROOT_DIR =
 let currentRoot =
     resolveRealDirectory(INITIAL_ROOT_DIR) ?? path.resolve(INITIAL_ROOT_DIR);
 
+function resolveCanonicalRoot(root: string): string | undefined {
+    const canonicalRoot = resolveRealDirectory(root);
+    return canonicalRoot !== undefined &&
+        path.relative(path.resolve(root), canonicalRoot) === ""
+        ? canonicalRoot
+        : undefined;
+}
+
 function getValidatedCurrentRoot(): string {
-    if (!isCanonicalDirectory(currentRoot)) {
+    if (resolveCanonicalRoot(currentRoot) === undefined) {
         throw new Error("The document root is no longer accessible");
     }
     return currentRoot;
@@ -921,11 +927,11 @@ app.post("/file/load", express.json(), (req: Request, res: Response) => {
             return;
         }
 
-        const resolvedPath = resolveExistingFileWithinRoot(
+        const resolvedPath = resolveWritableFileWithinRoot(
             getValidatedCurrentRoot(),
             newFilePath,
         );
-        if (resolvedPath === undefined) {
+        if (resolvedPath === undefined || !fs.existsSync(resolvedPath)) {
             res.status(403).json({
                 error: "Access to the file is forbidden or file not found",
             });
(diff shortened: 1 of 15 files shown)


ts/packages/agents/markdown/src/view/route/service.ts
```diff
@@ -18,12 +18,20 @@ import * as encoding from "lib0/encoding";
 import * as decoding from "lib0/decoding";
 import registerDebug from "debug";
 import sanitizeFilename from "sanitize-filename";
+import { randomUUID } from "node:crypto";
 import { isAllowedViewOrigin } from "./originAllowlist.js";
+import { resolvePathWithinRoot } from "./pathPolicy.js";
 import {
-    resolveExistingFileWithinRoot,
-    resolvePathWithinRoot,
```
jebrans, line 24: Can we rename to avoid confusion

ts/packages/agents/markdown/src/view/route/service.ts
```diff
@@ -849,11 +927,11 @@ app.post("/file/load", express.json(), (req: Request, res: Response) => {
             return;
         }
 
-        const resolvedPath = resolveExistingFileWithinRoot(
-            ROOT_DIR,
+        const resolvedPath = resolveWritableFileWithinRoot(
```
jebrans: Why did we switch?

ts/packages/agents/markdown/src/view/route/service.ts
```diff
@@ -1635,198 +1735,149 @@ Start typing to see the editor in action!
...
-                console.error(
-                    "[SSE] Failed to send operations to primary client:",
-                    error,
-                );
-                throw error;
-            }
-
-            // Send success confirmation back to agent
+            const documentId = path.basename(snapshot.boundRelativePath, ".md");
+            collaborationManager.setDocumentContent(
```
jebrans, line 1795: Will this replace the entire document even for a one word change?

ts/packages/agents/markdown/src/view/route/service.ts
```diff
@@ -1635,198 +1735,149 @@ Start typing to see the editor in action!
...
+                    revision: null,
                     timestamp: Date.now(),
                 });
+                return;
+            }
+            if (!snapshot.filePath || !snapshot.boundRelativePath) {
+                throw new Error("No markdown document is bound");
             }
-        })();
+            const document = readBoundDocument({
```
jebrans: Can this be async to handle large docs

jebrans (approved):

ts/packages/agents/markdown/src/view/route/service.ts
GeorgeNgMsft: I think this was a mistake, it was trying to re-use the helper it created

ts/packages/agents/markdown/src/view/route/service.ts
GeorgeNgMsft: Great catch, yeah, it's been updated to only edit target text properly

PR #2974, opened 2026-09-04
[Github Agent] add prFiles and prFailedChecks read actions

GeorgeNgMsft:
When a pull request is failing, you usually need two things before you can say anything useful about it: what the PR actually changed, and what the red checks are complaining about. Today the `github-cli` agent can tell you a PR exists and that its checks are red, but not much more, so answering either question means leaving the agent and running `gh` by hand.

This adds two read-only actions that answer those two questions directly.

`prFiles` lists the files a pull request changes, with each file's status and line counts, and optionally a short excerpt of each file's diff.

image

`prFailedChecks` explains why the checks are red. For each failing check it reports the workflow, timing, and a link, plus the error annotations GitHub recorded against it — usually the actual compiler or test error, with its file and line.

image

Highlights:

- **Deterministic fast actions.** Both are typed, read-only, and handled directly: they run `gh`, parse its JSON, and return a stable `rawData` payload. No reasoning step is involved, so an external MCP client can discover and call them on its own.
- **Bounded output, honestly labelled.** File counts, patch sizes, check counts, and annotation counts all have caps, and every result says explicitly what got truncated instead of quietly dropping it.
- **Diffs are opt-in.** When patches aren't requested, they're stripped server-side via `--jq`, so listing the files of a large PR doesn't download its whole diff.
- **Works on other repos, including by link.** Omit `repo` to use the current checkout, or pass an `OWNER/REPO` slug — or just the pull request's web link, which `gh --repo` itself rejects. The host is carried through, so GitHub Enterprise links work too.
- **Degrades instead of failing.** Missing logs, in-progress checks, checks from non-GitHub providers, forks, and permission errors each produce a clear partial result and a hint, rather than an error.
- **Natural language too.** Grammar rules are added so phrases like "what files did PR 123 change" and "why is PR 123 failing" work, but they're additive — the actions stand on their own.

Tests cover argument construction, JSON parsing and structured results, repo/link normalization, pagination and truncation boundaries, grammar, and the error paths.

### DIFF
--- a/ts/packages/agents/github-cli/src/github-cliActionHandler.ts
+++ b/ts/packages/agents/github-cli/src/github-cliActionHandler.ts
@@ -37,6 +37,8 @@ import {
     runSetupCommand,
     whichExists,
 } from "./setup.js";
+import { buildTableBlock } from "./structuredResults.js";
+import { GhResult, runPrFailedChecks, runPrFiles } from "./prDiagnostics.js";
 
 const execFileAsync = promisify(execFile);
 
@@ -360,6 +362,41 @@ async function runGh(args: string[], timeoutMs = 30_000): Promise<string> {
         windowsHide: true,
     });
     return stdout.trim();
+}
+
+// Run a gh CLI command and return its full result, including a non-zero exit
+// code, instead of throwing. Some gh commands report status through the exit
+// code while still writing the JSON we asked for (`gh pr checks` exits
+// non-zero when checks fail), and some failures should degrade a section of a
+// result rather than fail the whole action. `exitCode` is -1 when gh never ran
+// at all — a missing binary or a timeout.
+//
+// The buffer is larger than `runGh`'s because these commands can return whole
+// pull request diffs.
+async function runGhCapture(
+    args: string[],
+    timeoutMs = 60_000,
+): Promise<GhResult> {
+    try {
+        const { stdout, stderr } = await execFileAsync("gh", args, {
+            timeout: timeoutMs,
+            maxBuffer: 8 * 1024 * 1024,
+            windowsHide: true,
+        });
+        return { stdout, stderr, exitCode: 0 };
+    } catch (e) {
+        const err = e as {
+            stdout?: string;
+            stderr?: string;
+            code?: unknown;
+            message?: string;
+        };
+        return {
+            stdout: err.stdout ?? "",
+            stderr: err.stderr || err.message || "gh failed to run",
+            exitCode: typeof err.code === "number" ? err.code : -1,
+        };
+    }
 }
 
 // Sentinel values that mean "no assignee". `gh issue list --assignee <x>`
@@ -1258,13 +1295,9 @@ function makeStructuredTable<T>(
         pageSize?: number;
     },
 ): ActionResultSuccess {
-    const columns = colSpecs.map(({ value: _v, ...col }) => col);
-    const rows: TableCell[][] = objects.map((obj) =>
-        colSpecs.map((col) => col.value(obj)),
-    );
     // Cap long lists to a first page (client reveals the rest via "Show
     // more") unless the caller overrode it. All rows still ship.
-    const table: TableBlock = createTable(columns, rows, {
+    const table: TableBlock = buildTableBlock(colSpecs, objects, {
         pageSize: 15,
         ...tableOptions,
     });
@@ -1922,6 +1955,16 @@ async function executeAction(
         return validated.result;
     }
     action = validated.action;
+
+    // Multi-call read-only diagnostics. These compose several gh invocations
+    // into one structured result, so they run ahead of the single-command
+    // buildArgs/runGh path below.
+    if (action.actionName === "prFiles") {
+        return runPrFiles(action.parameters, runGhCapture);
+    }
+    if (action.actionName === "prFailedChecks") {
+        return runPrFailedChecks(action.parameters, runGhCapture);
+    }
 
     const args = buildArgs(action);
     if (!args) {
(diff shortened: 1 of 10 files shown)


robgruen:
If you're in here adding fun features I would ask you to add "review PR <blah> for me" and "investigate PR <blah> failure" and variations thereof. (although the latter requires copilot cli).

Image

Image

ts/packages/agents/github-cli/src/github-cliActionHandler.ts
```diff
@@ -362,6 +364,41 @@ async function runGh(args: string[], timeoutMs = 30_000): Promise<string> {
...
+// The buffer is larger than `runGh`'s because these commands can return whole
+// pull request diffs.
+async function runGhCapture(
+    args: string[],
+    timeoutMs = 60_000,
+): Promise<GhResult> {
+    try {
+        const { stdout, stderr } = await execFileAsync("gh", args, {
+            timeout: timeoutMs,
+            maxBuffer: 8 * 1024 * 1024,
```
robgruen, line 384: maybe a comment here on why we are doing 8MB

ts/packages/agents/github-cli/src/github-cliSchema.agr
```diff
@@ -270,6 +270,53 @@
     }
 };
 
+<PrFiles> = show files changed in PR $(number:number) -> {
+    actionName: "prFiles",
+    parameters: {
+        number
+    }
+}
+  | what files does PR $(number:number) change -> {
```
robgruen: "change", "touch", "edit", "modify"

ts/packages/agents/github-cli/src/github-cliSchema.agr
```diff
@@ -270,6 +270,53 @@
...
+    actionName: "prFiles",
+    parameters: {
+        number
+    }
+}
+  | show files changed in PR $(number:number) in $(repo:wildcard) -> {
```
robgruen, line 291: "show what's changed in ...."

ts/packages/agents/github-cli/src/github-cliSchema.agr
```diff
@@ -270,6 +270,53 @@
...
+    }
+};
+
+<PrFailedChecks> = show failing checks for PR $(number:number) -> {
+    actionName: "prFailedChecks",
+    parameters: {
+        number
+    }
+}
+  | why is CI failing on PR $(number:number) -> {
```
robgruen: why is (the) CI

ts/packages/agents/github-cli/src/github-cliSchema.ts
```diff
@@ -391,6 +393,69 @@ export type PrChecksAction = {
...
+
+        // OWNER/REPO slug (e.g. "microsoft/TypeAgent"). Omit to use the
+        // repository in the current directory.
+        repo?: string;
+
+        // Include an excerpt of each file's diff. Off by default because
+        // patches are large; turn it on to see the actual code changes.
+        includePatch?: boolean;
+
+        // How many files to return, newest API order. 1-300, default 50.
```
robgruen: is 50 the default that comes out of the tool itself?  If so no need to change...if you *want* it to be 50 then you could define: maxFiles: number | 50;

robgruen (approved):

ts/packages/agents/github-cli/src/github-cliSchema.ts
GeorgeNgMsft: That's fair, it's just arbitrary

ts/packages/agents/github-cli/src/github-cliSchema.agr
GeorgeNgMsft: Thanks, I'll add a bit more grammar too

ts/packages/agents/github-cli/src/github-cliActionHandler.ts
GeorgeNgMsft: It's just an arbitrary bound right now, but can leave a comment

GeorgeNgMsft:
> If you're in here adding fun features I would ask you to add "review PR for me" and "investigate PR failure" and variations thereof. (although the latter requires copilot cli).
> 
> Image Image

I can add this in a follow up PR later on, but right now I'm generally covering common Git related actions that are deterministic with no model dependency. Trying to make sure our list of _useful fast actions_ is well curated for the average dev scenario.

ts/packages/agents/github-cli/src/github-cliSchema.ts
GeorgeNgMsft: Kept `maxFiles` as an optional numeric override and documented why the implementation defaults to 50: it keeps the structured display manageable, while callers can explicitly request up to 300.

ts/packages/agents/github-cli/src/github-cliSchema.agr
GeorgeNgMsft: Added `change`, `touch`, `edit`, and `modify` as supported grammar alternatives, with coverage for each phrasing.

ts/packages/agents/github-cli/src/github-cliSchema.agr
GeorgeNgMsft: Added support and test coverage for both `why is CI failing` and `why is the CI failing`.

ts/packages/agents/github-cli/src/github-cliSchema.agr
GeorgeNgMsft: Added the `show what's changed in PR <number>` phrasing and a grammar test for it.

ts/packages/agents/github-cli/src/github-cliActionHandler.ts
GeorgeNgMsft: Documented why `runGhCapture` uses an 8 MiB buffer: GitHub returns patches before truncation, so the larger bounded buffer accommodates large PR responses and unusually large generated-file patches.

PR #2976, opened 2026-09-04
Add getGitDiff read action and fix code-agent error semantics

GeorgeNgMsft:
## What this does

Two small, related fixes to the Code Agent / Coda (VS Code extension bridge), both read-only:

1. **New `getGitDiff` action** — a deterministic, catalog-discoverable fast action for the `code` agent that returns the actual local Git diff: changed files plus bounded unified-diff patch/hunk text, split into staged and unstaged sections by default, or diffed against an explicit base ref (`base` parameter) when given. It selects a repository explicitly or falls back to the single open repo/workspace folder (`repository` parameter, for multi-root workspaces). Binary files, renames, copies, and deletions are reported with clear status/flags instead of garbled or dropped patch text, and output is bounded (per-file byte cap, per-section byte cap, file-count cap) with truncation flags so callers always get a well-formed, complete response instead of a runaway payload.
2. **Fixed error semantics** — when a Coda read action fails for an operational reason (e.g. no repo open, an unhandled action), the Code Agent now returns a *failed* `ActionResult` instead of a success-shaped result that merely happens to contain `{"error": ...}` in its text.

`getWorkspaceChanges` is unchanged.

### Why this matters for external MCP clients

`getGitDiff` needs no separate MCP wiring — it's defined the same way as every other code-agent action, so it's automatically visible through the existing active schema/catalog (the same pipeline `packages/actionSchema/test/regen.spec.ts` round-trips and `commandExecutor` reads for discovery). A new test (`codeActionsSchemaDiscovery.spec.ts`) proves it's present with a concise description and a stable, typed parameter list — deterministic and usable by an external tool-selecting client without going through TypeAgent's natural-language reasoning path.

## Highlights

- `packages/coda/src/gitDiffUtils.ts` (new) — pure, unit-tested diff parsing: quoted/unquoted Git header paths (including non-ASCII under both `core.quotePath` settings), trailing-tab disambiguation on spaced paths, binary/mode-only/rename/copy/deletion handling, and bounded truncation with reported caps.
- `packages/coda/src/handleReadActions.ts` — `getGitDiff` uses a single `repo.diffWith()` call per section (no serial per-file calls), filters output to paths inside an open workspace folder (checking both current and old paths for renames/copies), and reuses a shared `getGitApi()` helper.
- `packages/agents/code/src/codeActionsSchema.ts` — new `GetGitDiffAction` schema with a concise, standalone description and `base`/`repository` parameters.
- `packages/agents/code/src/codeActionHandler.ts` — operational failures (JSON `{error}` results, and the "action not handled" fallback) now produce a failed `ActionResult`.
- Tests: 28 new unit tests in `packages/coda/test/gitDiffUtils.spec.ts` (new `test:local` harness for coda, matching existing agent conventions) and 2 new suites in `packages/agents/code/test/` (`codeActionErrorSemantics.spec.ts`, `codeActionsSchemaDiscovery.spec.ts`) — 31/31 code-agent tests, 28/28 coda tests passing.
- All CLAUDE.md ratchet gates (lint, complexity, circular deps, tech debt) pass against `main`.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>


GeorgeNgMsft:
Closing for now, will reopen when ready for review

PR #2985, opened 2026-09-09
Update Copilot defaults to GPT-5.6

hillary-mutisya:
Switch Copilot chat and reasoning defaults to GPT-5.6 models, refresh fallback lists and docs, and improve model selection to prefer matching or nearby GPT tier variants.

### DIFF
--- a/ts/packages/aiclient/src/copilotModels.ts
+++ b/ts/packages/aiclient/src/copilotModels.ts
@@ -372,14 +372,21 @@ function isConcreteAvailableModel(model: ModelInfo): boolean {
     return model.id !== "auto" && model.policy?.state !== "disabled";
 }
 
-function versionedFamily(
-    modelId: string,
-): { family: string; version: number[] } | undefined {
-    const match = /^(.*?)-(\d+(?:\.\d+)*)$/.exec(modelId);
+type VersionedModelId = {
+    family: string;
+    version: number[];
+    tier: string | undefined;
+};
+
+function versionedFamily(modelId: string): VersionedModelId | undefined {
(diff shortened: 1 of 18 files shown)


ts/packages/dispatcher/dispatcher/src/reasoning/copilot.ts
```diff
@@ -178,7 +178,7 @@ async function sendMessageAndWaitWithCancellation(
     }
 }
 
-const FALLBACK_MODEL = "claude-opus-4.8";
+const FALLBACK_MODEL = "gpt-5.6-sol";
```
jebrans, line 181: https://github.com/microsoft/TypeAgent/blob/main/ts/config.defaults.yaml#L16 can you also update here

ts/packages/aiclient/src/copilotModels.ts
jebrans, line 483: Should we check this after the family matching logic that we have

ts/tools/scripts/generate-selfhost-config.mjs
```diff
@@ -35,7 +35,7 @@
  *   --force                     Overwrite an existing config.local.yaml.
  *   --ollama-host <url>         Ollama base URL (default http://localhost:11434).
  *   --chat-model <name>         Ollama chat model (default llama3.2).
- *   --copilot-model <name>      Copilot chat model (default claude-haiku-4.5).
+ *   --copilot-model <name>      Copilot chat model (default gpt-5.6-luna).
```
jebrans, line 38: One more place https://github.com/microsoft/TypeAgent/blob/main/ts/tools/scripts/install-typeagent.sh#L57 we can update the help text

jebrans (approved):

PR #2996, opened 2026-09-11
Fix pipeline packaging and stabilize smoke tests

hillary-mutisya:
Pin pnpm in generated bundles, isolate translation tests from PowerShell routing, and update Copilot tool tests for GPT-5.6 models.


GeorgeNgMsft (approved):
I see the tool test is deleted rather than updated to GPT 5.6, I’d assume that is because it’s flaky? If that is intentional, then this looks good to me. The PR says that it updated the tests though, so I wasn’t sure.

PR #3002, opened 2026-09-15
Add managed Copilot runtime provisioning

hillary-mutisya:
Install SDK-compatible Copilot runtimes from the authenticated TypeAgent feed, expose setup and status commands, and configure agent startup to use the managed executable. Update standalone and MSI installers to provision runtimes, launch interactive Copilot setup after installation, verify feed packages in publishing, and handle disabled plugin registrations.


ts/tools/scripts/install-typeagent.ps1
```diff
@@ -500,16 +590,10 @@ $pluginName = "typeagent"
...
-    if (-not (Test-Command npm)) {
-        Fail "GitHub Copilot CLI is required to register the plugin, and npm is not available to install it."
-    }
-
-    Write-Host "  Installing GitHub Copilot CLI (npm i -g @github/copilot)"
-    & npm install -g "@github/copilot"
-    if (-not (Test-Command copilot)) {
-        Fail "GitHub Copilot CLI was not found after installation."
-    }
+$pluginCopilotPath = if (Test-Command copilot) {
```
jebrans, line 593: We will revert to the old CoPilot. We are installing TypeAgent CoPilot version before this. Do we want that to be the default?

ts/tools/installers/wix/TypeAgent-AgentServer.wxs
jebrans, line 689: Should we re-oder this after the Copilot is installed in InstallPrereqs

ts/tools/scripts/test/copilotManagedRuntime.spec.mjs
```diff
@@ -0,0 +1,289 @@
...
+const testDir = path.dirname(fileURLToPath(import.meta.url));
+const tsRoot = path.resolve(testDir, "..", "..", "..");
+
+test("manifest records the resolved SDK-compatible Windows runtime", () => {
+    const manifest = createCopilotRuntimeManifest({
+        platform: "win32",
+        arch: "x64",
+    });
+    assert.equal(manifest.sdkPackage, "@github/copilot-sdk");
+    assert.equal(manifest.sdkVersion, "1.0.9");
```
jebrans, line 36: Will this fail if we bump the version?

jebrans (approved):

ts/tools/installers/wix/TypeAgent-AgentServer.wxs
hillary-mutisya: The copilot plugin is registered to work with the version of copilot CLI that the user has installed. That may be different from the SDK edition that TypeAgent is built against and runs against. For example, typeagent may build and run against version N but the user is running a more recent M>N version.

PR #3007, opened 2026-09-16
Close browser windows by title

datduyng:
The benchmark exposed wrong window-close routing, while source inspection found a separate execution bug that discarded `closeWindow.title`; this PR implements narrow title targeting, but it does not fix translation and provides no benchmark-score improvement.

## Before

```console
# Live-model check
$ agent-cli prompt 'Reply with exactly LIVE_MODEL_OK and nothing else.'
Sending prompt to GPT:
Reply with exactly LIVE_MODEL_OK and nothing else.

Full GPT response: (594.467541ms)
LIVE_MODEL_OK

$ agent-cli connect --memory --request "That's great. Also, close the entire browser window whose title bar says 'Quarterly Forecast Dashboard - Window 2'." --exit --port 9234
[🌐 browser] Translating 'That's great. Also, close the entire browser window whose title bar says 'Quarterly Forecast Dashboard - Window 2'.'
[🌐 browser] Executing action browser.closeAllWebPages
Agent 'browser' needs configuration before it can be used: The TypeAgent browser extension hasn't connected yet on this machine.
```

Separately, `browser.external.closeWindow({ title })` discarded `title`, and the extension closed whatever window `chrome.windows.getCurrent()` returned.

## E2e testing

```
@action browser.external closeWindow --parameters '{"title":"Quarterly Forecast Dashboard - Window 2"}'
```
Below show a demo that when there are 2 windows - close Window with title `Quarterly Forecast Dashboard - Window 2` - only `Quarterly Forecast Dashboard - Window 2` closed despite the focused window is on Operations Dashboard`
CleanShot 2026-09-16 at 13 32 50@2x

## After

The live command above still selects `browser.closeAllWebPages`; it demonstrates the remaining translation gap, not the new title-targeting behavior.

The implementation now forwards `title`, closes exactly one normal window whose active-tab title matches case-insensitively, and returns an action error without closing anything on zero or multiple matches. Untitled requests still close the current window. The inline shell browser remains unsupported.


ts/packages/agents/browser/src/agent/externalBrowserActionSchema.mts
```diff
@@ -33,7 +33,7 @@ export type CloseTab = {
 export type CloseWindow = {
     actionName: "closeWindow";
     parameters: {
-        title?: string;
+        title?: string; // case-insensitive exact title of the active tab in the target window, ignoring surrounding whitespace; omit to close the current window
```
robgruen: "case-insensitive" and "exact title" are mutually exclusive so I would either fix the logic or the comment.

ts/packages/agents/browserExtension/src/extension/serviceWorker/externalBrowserControlServer.ts
```diff
@@ -33,6 +33,33 @@ async function ensureActiveTab() {
     return targetTab;
 }
 
+async function getWindowIdByTitle(title: string): Promise<number> {
+    const normalizedTitle = title.trim().toLowerCase();
```
robgruen: I would prefer if we make this a setting on whether or not we automatically lower case the title.  I could see this causing trouble at some point.  I think it's ok to default to do this but we need a way to turn it off as well.

ts/packages/agents/browserExtension/src/extension/serviceWorker/externalBrowserControlServer.ts
```diff
@@ -33,6 +33,33 @@ async function ensureActiveTab() {
...
+    const windows = await chrome.windows.getAll({ populate: true });
+    const matchingWindows = windows.filter((window) => {
+        if (window.type !== "normal" || window.id === undefined) {
+            return false;
+        }
+        const activeTab = window.tabs?.find((tab) => tab.active);
+        return activeTab?.title?.trim().toLowerCase() === normalizedTitle;
+    });
+
+    if (matchingWindows.length === 0) {
```
robgruen, line 66: this will only work if it's the active tab, is that intentional?  what if you want to close an inactive tab?

robgruen (approved):

ts/packages/agents/browser/src/agent/externalBrowserActionSchema.mts
datduyng: Fixed the wording. It now describes matching the full trimmed active-tab title, with case-insensitive matching as the default and a configurable case-sensitive mode.

ts/packages/agents/browserExtension/src/extension/serviceWorker/externalBrowserControlServer.ts
datduyng: Added `caseInsensitiveWindowTitleMatching` under Extension Options > Browser Control. It defaults on, persists in `chrome.storage.sync`, and is read at match time; turning it off requires matching capitalization.

ts/packages/agents/browserExtension/src/extension/serviceWorker/externalBrowserControlServer.ts
datduyng: Yes, active-tab-only is intentional: a browser window’s visible title comes from its active tab. Inactive-tab titles identify tabs within that window, not separate windows. I added a code comment to make this explicit.

PR #3011, opened 2026-09-16
Domnguyen/seal tools evaluation

datduyng:


ts/packages/benchmarks/src/translationBench/public_datasets/Seal-Tools/toTypeAgentSchema.ts
```diff
@@ -0,0 +1,347 @@
...
+        canonicalPayloadHash: sha256(canonical),
+        transformVersion: 1,
+    };
+
+    return applySealToolsTypeAgentOverride({
+        id: `sealtools-${row.id}`,
+        utterance: parsed.utterance,
+        schemaName: SEAL_SCHEMA_NAME,
+        tools,
+        sealToolsGoldActions: parsed.calls.map((call) =>
```
GeorgeNgMsft, line 312: The number conversion below outputs the number as a string rather than number stored in a plain object `{ age: { __pythonNumber: "44" } }`. However, the scorer seems to expect a PythonNumber `{ age: new PythonNumber("44") }` which may not be compared and scored. Double check that this isn't problematic, if it is, then you may want to update the scorer too so that it matches.

ts/packages/benchmarks/src/translationBench/public_datasets/Seal-Tools/toTypeAgentSchema.ts
```diff
@@ -0,0 +1,347 @@
...
+    });
+    const canonicalPayloadHash = sha256(
+        JSON.stringify({
+            utterance: row.utterance,
+            expectedActions,
+            order: row.order,
+        }),
+    );
+    return {
+        ...row,
```
GeorgeNgMsft, line 234: Row's dimensions may need updating, since expectedActions arity can be changed during override, the dimensions.arity property in row also needs to be updated so that the values match.

GeorgeNgMsft (approved):
Two minor issues to address

ts/packages/benchmarks/src/translationBench/public_datasets/Seal-Tools/toTypeAgentSchema.ts
datduyng: Fixed in d23e0d251. The Seal-compatible scorer now recognizes the serialized __pythonNumber marker (including nested values and case-insensitive comparison), so JSONL round-tripped gold compares with PythonNumber predictions. Added a converter-to-scorer regression test.

ts/packages/benchmarks/src/translationBench/public_datasets/Seal-Tools/toTypeAgentSchema.ts
datduyng: Fixed in d23e0d251. Overrides now recompute dimensions.arity and dimensions.shape from corrected expectedActions; a regression test covers a two-action source reduced to one corrected action.

PR #3013, opened 2026-09-16
Use GPT-5.6 Luna for TypeAgent GPT-4o paths

datduyng:
Did verification test and put it in internal Loop


datduyng:
Remove to put to Loop

hillary-mutisya (approved):

robgruen (changes requested):
see my private message

ts/packages/aiclient/src/providerMode.ts
```diff
@@ -85,6 +86,7 @@ const COPILOT_MAP: Record<CanonicalName, string> = {
     GPT_5: "gpt-5.6-sol",
     GPT_5_MINI: "gpt-5.6-terra",
     GPT_5_NANO: "gpt-5.6-luna",
```
robgruen, line 88: GPT_5_NANO != gpt-5.6-luna.  Can you please fix same thing for GPT_5_MINI and GPT_V.  If the goal is to have  "cheap", "normal", "expensive" models then we should consider using that nomenclature.  Here the name of the model implies the model being used and having these not match can lead to more or less cost burden on users so we shouldn't false advertise.  

Give you can say "switch the model to GPT_4" it should switch to 4 not something named 4 that it something else underneath.

ts/packages/dispatcher/dispatcher/src/context/session.ts
```diff
@@ -414,11 +414,8 @@ const defaultSessionConfig: SessionConfig = {
     request: DispatcherName,
     translation: {
         enabled: true,
-        // Default translation model. "GPT_4_1" resolves to the gpt-4.1
-        // deployment in both config.local.yaml and the CI build-pipeline-kv
-        // config, keeping local dev and CI on the same model. Override per
-        // session with `@config translation model <name>`.
-        model: "GPT_4_1",
+        // Override per session with `@config translation model <name>`.
```
robgruen, line 417: This assumes that the config file being used has a "GPT_5_6_LUNA" endpoint in it.  Previously it was guaranteed that GPT_4_1 was in everyone's config file...any partner team that is currently using TypeAgent will break if they sync with the repo and don't update their config file. If we supplied their config file or they don't have access to our keyvault to get keys they will be in an unrecoverable state.

I would consider doing a fallback to GPT_4_1 if there's no endpoint for the one being supplied here.

ts/packages/aiclient/src/providerMode.ts
robgruen: If you want to leave it as is we should at least put some comments in there stating why there's a mismatch between the name and the actual model being used.

robgruen (changes requested):
Let's at a minimum fix the default model logic with a fallback otherwise we might break some existing customers.

ts/packages/dispatcher/dispatcher/src/context/session.ts
datduyng: btw, I put the fallback in translateRequest because `session.ts` has no idea whether an endpoint exists. The session config is a plain data object

robgruen (approved):

ts/packages/aiclient/src/providerMode.ts
datduyng: Followed up offline - will make new pr to address this

PR #3017, opened 2026-09-17
Add translation benchmark policy generation

datduyng:
Adds generated action policy and eligibility data so translation benchmarks can use a reviewed, reproducible action set; this does not claim a benchmark score change.

```console
$ node_modules/.bin/tsc -b && node --no-warnings --experimental-vm-modules node_modules/jest/bin/jest.js --testPathPattern='translationBench\.(actionQualityPicker|policy|policyGenerator)[.]spec[.]js' --runInBand
Before: Policy generation, eligibility selection, and their tests were not available.
After:  3 test suites passed, 64 tests passed, 0 failed.
Result: The benchmarks package compiled and all focused policy tests passed.
```


datduyng:
Replaced by #3020, #3021, and #3022. Each replacement is independently based on main.

PR #3018, opened 2026-09-17
Add ambiguity checks to benchmark generation

datduyng:
Adds multi-model ambiguity checks to reject unclear generated benchmark cases before approval; this does not claim a benchmark score change.

```console
$ node_modules/.bin/tsc -b && node --no-warnings --experimental-vm-modules node_modules/jest/bin/jest.js --testPathPattern='translationBench[.]ambiguityProbe[.]spec[.]js' --runInBand
Before: Generated cases had no ambiguity-probe classification or focused probe tests.
After:  1 test suite passed, 6 tests passed, 0 failed.
Result: The benchmarks package compiled and all focused ambiguity tests passed.
```


GeorgeNgMsft (approved):

ts/packages/benchmarks/src/translationBench/synthesizer/synthesizerPrompts.ts
```diff
@@ -399,7 +421,7 @@ export function loadTranslationBenchParameterGraderPromptPack(
     return loadPack(
         PARAMETER_GRADER_PROMPT_FILE,
         translationBenchParameterGraderPromptPackSchema,
-        promptsDir,
+        promptsDir ?? TRANSLATION_BENCH_POLICY_PROMPTS_DIR,
```
GeorgeNgMsft: This change was flagged as potentially incomplete, it seems like the expected file in the TRANSLATION_BENCH_POLICY_PROMPTS_DIR default doesn't seem to exist? You should double check if the prompt file needs to be moved here.

ts/packages/benchmarks/src/translationBench/synthesizer/ambiguityProbe.ts
```diff
@@ -236,3 +252,385 @@ export function translationBenchAmbiguityCasesClear(
...
+                    code: "OTHER",
+                    path: "$",
+                    message: "Ambiguity probe found no positive utterances",
+                    suggestedFix: "Ensure seed is a positive gold label",
+                },
+            ],
+        };
+    }
+
+    if (translationBenchAmbiguityCasesClear(cases)) {
```
GeorgeNgMsft, line 553: It seems like the shortcut here may result in failed cases being ignored and so multi-model checks will be treated as a unanimous success. The original logic for the classifier used seems to ignore failed calls which was fine for it's original usage. However, the classifier seems to be reused for this and it ignores model failures, so it's reporting incorrectly here. Please review the logic here

ts/packages/benchmarks/src/translationBench/synthesizer/synthesizerPrompts.ts
datduyng: Reverted - the default now loads from the synthesizer prompts dir where parameter-grader.prompt.yaml actually lives; the policy/ constant is removed.

ts/packages/benchmarks/src/translationBench/synthesizer/ambiguityProbe.ts
datduyng: Fixed - added a partial_errors agreement state. Failed probes are no longer dropped before scoring, so 2 errors + 1 gold is not unanimous. It now fails closed with a deterministic issue ("N of M model(s) failed"), and a regression test covers the survivor-agreement case.

PR #3021, opened 2026-09-17
Add benchmark parameter policy generation

datduyng:
Adds parameter-policy generation, schema conversion, prompt assets, and a dedicated CLI while leaving catalog generation and eligible-action selection unchanged.

```console
$ node_modules/.bin/tsc -b && node --no-warnings --experimental-vm-modules node_modules/jest/bin/jest.js --testPathPattern='translationBench[.]policyGenerator[.]spec[.]js' --runInBand
Before: The benchmarks package had no policy-based parameter generator or gen-policy command.
After:  47 focused generator tests passed, 0 failed.
Result: The benchmarks package compiled and the parameter-policy tests passed.
```


ts/packages/benchmarks/src/translationBench/policy/policyGenerator.ts
```diff
@@ -0,0 +1,2317 @@
...
+        throw new Error(
+            `Invalid field grader for ${actionIdLabel}.${fieldName}: rule`,
+        );
+    }
+    if (LEGACY_RULE_RE.test(field.rule) || /default/i.test(field.rule)) {
+        throw new Error(
+            `Invalid field grader for ${actionIdLabel}.${fieldName}: legacy/default rule '${field.rule}'`,
+        );
+    }
+    if (field.source !== "hardcode" && field.source !== "llm") {
```
jebrans: Will this fail for regex

ts/packages/benchmarks/src/translationBench/policy/index.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export * from "./paramTypes.js";
```
jebrans: Some of the files are similar to files in catalogGenerator folder

jebrans (approved):

ts/packages/benchmarks/src/translationBench/policy/policyGenerator.ts
datduyng: No - the loader accepts both: `ActionParamClassifySource` is `"regex" | "hardcode" | "llm"` and the validators accept legacy `"regex"` values on load. Newly generated entries use `"hardcode"`.

ts/packages/benchmarks/src/translationBench/policy/index.ts
datduyng: Fixed - the forked copies are gone. `policyGenerator.ts`, `paramTypes.ts`, `schemaTypeConvert.ts`, and the duplicate prompt yaml were merged back into `synthesizer/catalogGenerator/` (now split into `paramGraderTypes` / `paramGraderClassify` / `paramGraderValidate` / `paramGraderNameHeuristics`). `policy/` keeps only policy-specific files.

PR #3023, opened 2026-09-17
Report agent server startup failures accurately

jebrans:
## Summary

- Report the actual agent-server startup outcome instead of treating every readiness failure as missing provisioning.
- Inspect the resolved local config only after startup fails and distinguish missing, invalid, structurally valid, and unavailable states.
- Capture daemon output in a bounded per-user log by default while keeping raw diagnostics out of automatic terminal output.
- Add regression coverage for early exits, timeouts, spawn failures, occupied ports, logging failures, rotation, config inspection, and Windows path casing.

## Root cause

`typeagent-serve.mjs start` waited only for port 8999 and printed one static provisioning recommendation whenever readiness failed. It did not inspect `config.local.yaml`, observe daemon exit state, or retain normal startup output. On Windows, the non-debug path also tracked a short-lived `cmd.exe start` wrapper instead of the actual daemon process.

## Fix

- Add a side-effect-free `inspectConfigFile()` API to `@typeagent/config` using the existing YAML validation and flattening rules.
- Spawn the real Node daemon directly with detached, hidden process options and track spawn errors and exit code/signal.
- Race readiness polling against daemon lifecycle so early failures return immediately and live-process timeouts remain distinct.
- Recommend provisioning only when the resolved local config file is actually missing.
- Write stdout/stderr to an owner-restricted log, rotate it above 1 MiB, and retain one previous log.
- Keep raw daemon output and parser details out of automatic startup errors; expose the log only through its path or the explicit `logs` command.
- Preserve occupied-port idempotence while avoiding the unsupported claim that any listener is definitely TypeAgent.

## Tests

- `@typeagent/config` TypeScript build.
- 29 config loader tests.
- 14 focused `typeagent-serve` tests.
- Full `tools/scripts/test/*.spec.mjs` suite.
- Repository-pinned Prettier check.
- Lint, complexity, circular-dependency, and test-debt PR ratchets.
- Deployed Windows artifact smoke test from a path containing spaces, including an immediate exit with code 17, config inspection through the deployed dependency, private log capture, and terminal redaction.

## Rollout

Keep this PR in draft for owner review. No migration or configuration rewrite is required. Existing successful starts, tunnel opt-in behavior, and occupied-port return code remain unchanged.

## Rollback

Revert the commit. The previous static readiness failure message and debug-only daemon logging behavior will be restored; no persisted data migration is involved.


jebrans:
Replaced by #3033, which uses the same head commit from an upstream microsoft/TypeAgent branch so the trusted TypeAgent smoke tests can run.

PR #3025, opened 2026-09-17
Fix CLI interaction input routing

jebrans:
## Summary
- suspend the CLI raw command prompt while a server-pushed interaction is waiting for input
- restore the prompt after the interaction completes without leaking approval input into a queued command
- add a regression test for permission input while the normal command prompt is active

## Root cause
The normal raw-mode prompt and a server-pushed permission prompt both attached stdin data listeners. A selection such as `1` was delivered to both, producing duplicated input and submitting the selection as a separate command. That command could cancel the active reasoning request.

## Validation
- `git diff --check` passed
- Not run: this checkout has no `ts/.npmrc` or dependencies. Provisioning via `node tools/scripts/setupWorktree.mjs` was blocked by Azure Key Vault HTTP 403 while retrieving the private npm configuration.


hillary-mutisya (approved):

jebrans:
Closing this fork-backed PR because GitHub does not allow changing an existing pull request head repository. The fix has been pushed directly to `microsoft/TypeAgent:dev/jebransyed/fix-cli-interaction-input`; a replacement draft PR will be opened from that upstream branch.

PR #3026, opened 2026-09-17
Fix Markdown model resolution and document update targeting

GeorgeNgMsft:
Fix Markdown updates failing with a missing Azure endpoint or reporting success after editing the wrong document. The agent now resolves its model through the shared AI client and explicitly selects files named in update requests, rather than silently updating the session's `live.md` after a restart.

- Use the canonical `GPT_4_O` model name and shared provider/endpoint-pool resolution for normal and editor streaming updates.
- Add `documentPath` to both update actions, supporting relative paths and absolute paths within the working directory while rejecting missing or out-of-root files.
- Include the destination in update results and report no-op responses without claiming changes were applied.
- Honor streaming model `Result<string>` responses and surface model failures instead of fabricating successful placeholder content.
- Clarify raw Markdown offsets and append positions to preserve existing content.
- Add regression coverage for configuration, restart-style target selection, persistence, errors, and no-op results; document model and target behavior.

Validation: 99 tests passed across nine source-backed Markdown suites, plus a live model run covering natural-language action translation and persisted edits to an explicitly named file while leaving the session document unchanged.


datduyng (approved):
UI only change - giving quick approval as we can always revert if things break

GeorgeNgMsft:
> UI only change - giving quick approval as we can always revert if things break

It's just a small bug fix

PR #3028, opened 2026-09-18
Avoid repeated temporary output in CLI

jebrans:
## Summary

- Stop the connected CLI from printing every cumulative Copilot reasoning snapshot as a separate line.
- Preserve inline-to-block newline handling when temporary output is skipped.
- Suppress source headers for agent-initiated temporary messages.

## Root cause

The Copilot reasoning loop sends cumulative progress snapshots through `appendDisplay(..., "temporary")`. The display contract defines temporary content as replaceable by the next message, but the enhanced CLI treated every append mode except `inline` as a permanent block. Each cumulative snapshot was therefore committed to terminal scrollback.

## Fix

Filter temporary messages at the start of the enhanced CLI's `renderAgentMessage` path. The request spinner remains visible while temporary content is suppressed, and permanent `inline`, `block`, `step`, and `setDisplay` output continues through the existing renderer.

Filtering before rendering also ensures skipped temporary messages do not mutate `lastAppendMode` or emit an orphan agent source header.

## Tests

- Dependency-aware TypeScript build: `pnpm run build agent-cli` (296 tasks passed).
- CLI queueing suite: 27 tests passed.
- Repository-pinned Prettier check passed.
- Lint, complexity, circular-dependency, and test-debt PR ratchets passed against `origin/main`.

The full CLI ESM run also passed the changed queueing suite but currently reports 13 unrelated failures in the existing conversation and interaction suites on `main`. The package's bare `pnpm --filter agent-cli test` command additionally omits Node's required `--experimental-vm-modules` flag on Windows.

## Independent review

A Gemini subagent reviewed the initial patch and identified two valid edge cases: preserving the prior inline render state and filtering agent source headers before output. Both were addressed with regression tests. A second review of the final commit found no blocking, regression, or security issues. A suggestion to modify the dispatcher's separate basic console helper was intentionally excluded because the reported connected CLI path uses `createEnhancedClientIO`.

## Risk

Low. The behavior change is limited to `temporary` messages in the connected CLI. These messages are transient by contract and are already replaced rather than retained by richer clients. Users continue to see the processing spinner and all permanent output.

## Rollback

Revert the commit to restore the previous behavior of printing temporary snapshots into CLI scrollback.


jebrans:
Independent review follow-up:

- Accepted: preserve `lastAppendMode` when temporary output is skipped. The filter now runs before `displayContent`, with an `inline -> temporary -> block` regression test.
- Accepted: prevent orphan `[source]` headers for agent-initiated temporary messages. The filter now runs before source-header rendering, with dedicated coverage.
- Not applied: changing the dispatcher's separate `createConsoleClientIO` helper. The reported connected CLI flow uses `createEnhancedClientIO`; widening this PR would affect a separate execution path without evidence from this bug.
- Final Gemini re-review found no remaining blocking, regression, or security findings.

Validation after the review changes: dependency-aware build passed, all 27 queueing tests passed, formatting passed, and all four PR ratchets passed.

hillary-mutisya (approved):

jebrans:
Superseded by #3032. The replacement uses the identical commit from microsoft/TypeAgent:dev/jebransyed/fix-cli-temporary-output so the Azure smoke-test pipeline has the required repository permissions.

PR #3030, opened 2026-09-18
Add text-embedding-3-small support

hillary-mutisya:
Add text-embedding-3-small support through Copilot’s endpoint. Configure Copilot embeddings by default in installers and provisioning, while preserving explicit local overrides. Update @config modelProvider copilot to switch both chat and embedding providers and rebuild semantic routing with the new model. Include transport, configuration, dispatcher, and installer coverage.


ts/packages/dispatcher/dispatcher/src/context/appAgentManager.ts
```diff
@@ -794,6 +794,38 @@ export class AppAgentManager
...
+                } catch (e) {
+                    debugError(
+                        `Failed to rebuild embeddings for ${config.schemaName}: ${e}`,
+                    );
+                }
+            }
+            await Promise.all(pending);
+        }
+
+        this.actionSemanticMap = nextMap;
```
jebrans, line 826: This values is set even when there is failure in addActionSchemaFile Should we preseve the old map on failures?

jebrans (approved):

ts/packages/dispatcher/dispatcher/src/context/appAgentManager.ts
hillary-mutisya: No. Rebuild is called when there is some reason why we want to replace the previous state e.g. if state is corrupt, or if items have been updated such that their embeddings are no longer valid. If rebuild fails, the nextMap is set to ` new ActionSchemaSemanticMap();`. Having a new map is better than keeping the possibly corrupt previous map that users asked us to rebuild

PR #3035, opened 2026-09-18
Migrate TypeAgent plugin capture to Copilot extensions

hillary-mutisya:
Replace legacy command hooks and transcript parsing with typed Copilot SDK session events. Add native TypeAgent commands, PowerShell guidance, macro capture, history tracking, and extension artifact staging.

Add focused lifecycle, trace, command, and artifact tests.


datduyng (approved):
Approve. Migration code is verifiable through unit test which can easily revert if something go wrong

PR #3045, opened 2026-09-21
Optimize browser durable-memory ingestion

hillary-mutisya:
Move browser imports to a single durable-memory extraction pipeline and make indexing incremental for newly added sources.

- Remove duplicate browser-side knowledge extraction.
- Append new documents to copied index generations instead of rebuilding the entire corpus.
- Reduce fetched HTML before Markdown conversion.
- Suppress URL and link entities that add little retrieval value.
- Aggregate structural chunks using a safe 7,500-token limit.
- Increase extraction concurrency from 4 to 8.
- Increase embedding batch size from 8 to 64.
- Add structured ingestion stage and timing traces.
- Share the production durable-memory factory with live acceptance tests.
- Preserve atomic publication, restart durability, and full rebuild behavior for source replacement and corpus clearing.

Full-page Luna import latency decreased from about 450 seconds to 34-54 seconds while retaining expected entities, topics, and search results.

Validation:
- Browser: 249 tests passed
- Memory service: 12 tests passed
- KnowPro: 48 tests passed
- Production incremental acceptance passed with real Luna and embeddings
- Build, formatting, diagnostics, and repository quality gates passed


datduyng (approved):

datduyng:
Since the PR is about restructuring, unit testing and e2e testing are sufficient

